PR #242, opened 2026-05-05
helm: align chart with bow-config schema and harden secret handling

yochze:
- fix auth.mode rendering at top level (was nested under features:, silently dropped)
- remove duplicate features: block that was clobbering allow_* feature flags
- add LDAP / Active Directory support (config.ldap.* with bind_password Secret-only)
- add OIDC group-sync fields (label, icon, sync_groups, group_claim, resolve_group_names) required for Microsoft Entra installs
- gate sensitive ConfigMap envs (encryptionKey, googleClientSecret, smtp.password, licenseKey, googleClientId) on hasPrefix "\${" so placeholder values don't land in the ConfigMap as literal strings; real values come from config.secretRef
- drop dead BOW_OIDC_JSON env var (never read by backend, leaked oidc secrets)
- clean up SMTP gate: enabled or host implies on; fail if enabled without host
- add NOTES.txt with plaintext-secret detection and missing-encryption-key warning
- bump chart 1.0.14 -> 1.1.0 (additive: new ldap block, new oidc fields, conditional secret emission; keeps backward compat for plaintext values)
- README: new Secrets handling section, Entra and LDAP install examples

### DIFF
--- a/k8s/AGENTS.md
+++ b/k8s/AGENTS.md
@@ -37,11 +37,25 @@ Concise overview of `@k8s/` with emphasis on the Helm chart layout, how values m
   - `postgresql.auth.existingSecret`: Use an existing secret instead of inline values.
 - **App configuration & secrets**
   - `config.secretRef`: Name of an existing `Secret` whose keys override defaults from the `ConfigMap`.
-  - Environment keys commonly used by the app (set via Secret):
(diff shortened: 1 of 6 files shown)


Dimss (approved):
looks good to me

PR #516, opened 2026-07-02
[codex] Speed up pending instruction checks

yochze:
## Summary

This speeds up the instruction pending-change path that was making `/agents` and instruction routes slow.

Changes:
- skip inherited build snapshot rows whose proposed instruction version is identical to the base build version before doing text diff work
- add an equality fast path in `compute_hunks()` so identical strings never enter `difflib.SequenceMatcher`
- reuse the live pending id set inside one instruction-list request instead of recomputing it while filtering and decorating rows

## Root Cause

Pending builds snapshot all instructions, so most `BuildContent` rows are carry-over rows rather than real edits. The previous code still ran word-level hunks over those rows, including many identical strings. With whitespace tokenization and `SequenceMatcher(autojunk=False)`, identical large texts were surprisingly expensive and amplified across build snapshots.

## Validation

- `python3 -m py_compile backend/app/services/text_hunks.py backend/app/services/instruction_service.py`
- hot-patched the running `bow-app` container, compiled inside `/opt/venv`, restarted, and confirmed Docker health returned healthy
- production probes after patch:
  - `/api/instructions/counts`: ~0.69-0.72s
  - `/api/instructions?limit=50`: ~0.71s TTFB, ~1.02s total
  - `/api/instructions?limit=50&pending_only=true`: ~0.69s TTFB, ~0.84s total
  - `/api/instructions/pending-changes`: ~0.63-0.66s

For comparison, the slow path before the fix spent ~2.43s in Python diffing for the pending sweep on this dataset, and list/pending routes were previously several seconds.


yochze:
Closing this draft because its head branch picked up unrelated commits after creation. Replaced by #517, which contains only the pending-instruction speedup.

PR #735, opened 2026-07-20
fix(prompt): stop first home-page message failing with HTTP 400 when Auto model is selected

yochze:
**What:** The home-page prompt box forwarded the `'auto'` model sentinel verbatim into the new report's URL, so the first completion was submitted with `prompt.model_id='auto'` and the backend rejected it with **HTTP 400**.
**Why it existed:** the in-report submit path already maps `'auto' → null` (`modelIdForPayload`), but `createReport()` used the raw `selectedModel` — the two payload builders drifted.

## Executive summary (customer altitude)

A user opens Bag of Words, types their first question on the home screen, and instead of an answer they get a red **`Error: HTTP error! status: 400`**. It only clears if they happen to type a *second* message. For a brand-new user or a fresh conversation this is the first thing they see — it reads as "the product is broken." It triggers specifically when the org's **Auto model router** is on (the default model chip shows "Auto"), which is the recommended setup, so it hits exactly the orgs following our guidance.

## 4a. Product/UX impact (before → after)

| Step | Before (bug) | After (fix) |
|---|---|---|
| Type "hi" on home, submit | navigates to `…?model_id=auto` | navigates to `…?model_id=` (empty) |
| First assistant reply | 🔴 `Error: HTTP error! status: 400` | 🟢 real greeting streams |
| To get an answer | user must send a 2nd message | works on the 1st message |

Evidence: `media/pr/promptboxv2-home-error/before/2-report-first-message.png` (400) vs `.../after/3-first-message-success.png` (greeting).

## 4b. Technical mechanism (before → after)

```mermaid
flowchart LR
    H["Home box · Auto selected<br/>selectedModel='auto'"]
    subgraph BUG["Before"]
      H --> R1["createReport()<br/>model_id: selectedModel → ?model_id=auto"] --> P1["report page submits<br/>prompt.model_id='auto'"] --> B1["get_model_by_id('auto') → None<br/>🔴 HTTP 400"]
    end
    subgraph FIX["After"]
      H --> R2["createReport()<br/>model_id: modelIdForPayload → ?model_id="] --> P2["report page submits<br/>prompt.model_id=null"] --> B2["router/default resolves<br/>🟢 200, streams"]
    end
```

## Reviewer decision box

| | |
|---|---|
| **Risk** | Low — one-line frontend change, aligns two payload builders |
| **Blast radius** | Home-page "start a new chat" flow only; in-report submit unchanged |
| **Behavior change (intended)** | First home-page completion now sends the mapped model id (`auto → null`) instead of the raw sentinel |
| **Verified ✅** | Live UI reproduce→fix→verify (real Anthropic model); new e2e contract test (4 passed) |
| **NOT verified ⚠️** | Non-Auto/pinned-model home submits were already correct and were not re-exercised; no CI run observed on this branch yet |
| **Look first** | `frontend/components/prompt/PromptBoxV2.vue` `createReport()` |

## Evidence / KPI

| Check | Result |
|---|---|
| New e2e test `tests/e2e/test_completion_model_id_resolution.py` | **4 passed** (sqlite) |
| Live harness `tools/agent/repro_home_auto_400.mjs` — before fix | `REPRO CONFIRMED` (first POST completions → 400) |
| Live harness — after fix | `FIX VERIFIED` (200; timeline `user:success, system:success`) |
| Failures attributable to THIS PR | none observed |

## Context map

- **Key files:** `frontend/components/prompt/PromptBoxV2.vue` (fix, `createReport()`); `frontend/pages/reports/[id]/index.vue:4623` (reads `?model_id`); `backend/app/services/completion_service.py:272,2026` (resolution + 400); `backend/app/services/llm_service.py:212` (`get_model_by_id`).
- **Convention this establishes:** any code that emits a model id must send `modelIdForPayload` (sentinel-mapped), never the raw `selectedModel`.
- **Artifacts:** `docs/feedback-loops/promptboxv2-home-error.md`, `tools/agent/repro_home_auto_400.mjs`, `backend/tests/e2e/test_completion_model_id_resolution.py`, before/after screenshots under `media/pr/promptboxv2-home-error/`.

## Deep detail + honesty

- **Root cause:** `_resolve_completion_models` treats a non-empty `prompt.model_id` as an explicit pick (`completion_service.py:272`); `get_model_by_id('auto')` finds no such model and returns `None`, so `create_completion_stream` raises 400. The frontend fix keeps the sentinel from ever leaving the client.
- **Road not taken:** teaching the backend to treat `'auto'` as "use router/default." Rejected — `'auto'` is a frontend-only sentinel; accepting it server-side blurs the contract and would mask other bad-id callers. The e2e test deliberately keeps *any* unknown id a 400.
- **Regression guard scope:** the new e2e test guards the backend contract (unknown id → 400; omitted → default resolves); it passes both before and after the frontend change since the backend was never wrong. The frontend regression itself is guarded by the runnable `repro_home_auto_400.mjs` loop.
- **Unrelated pre-existing bug hit while seeding:** `POST /api/llm/models` 500s (`LLMService.create_model` missing — also noted in `promptboxv2-queue-steering.md`). Not touched here; models for the live loop were seeded directly.

https://claude.ai/code/session_01NSaVdH4hvLNQrGqirTQT2E

---
_Generated by [Claude Code](https://claude.ai/code/session_01NSaVdH4hvLNQrGqirTQT2E)_


yochze:
Added a second, related fix in `3168686` (same Auto-model root cause family).

**Symptom:** the LLM provider logo on assistant messages was missing on freshly-streamed replies — it only reappeared after a full page reload.

**Cause:** the avatar badge renders only when the message has a `model` (`frontend/pages/reports/[id]/index.vue:269`). The optimistic assistant bubble is created with `model: data.model_id || undefined`; with **Auto** selected `data.model_id` is `null`, so the live message had no model → no badge. `[DONE]` calls `loadReport()`, which refreshes report metadata only (not `messages`), so the badge stayed blank until reload.

**Fix:** the backend stamps the resolved model onto the `completion.started` SSE event (`completion_service.py:2409`), and the report page applies it to the live bubble in the `completion.started` handler. Badge now shows live.

**Verification:**
- e2e guard `test_completion_started_event_carries_resolved_model` — asserts `completion.started` carries a non-empty resolved model (5 passed total in that file).
- Live Playwright before/after (`tools/agent/verify_live_model_badge.mjs`, fresh report, Auto, no reload): **0 → 1** provider badge on the live message.

Docs: extended `docs/feedback-loops/promptboxv2-home-error.md` with this loop.

---

PR #789, opened 2026-07-26
Stop SharePoint tests and OneDrive sign-in from walking the whole drive

yochze:
Neither delay was the connection check itself.

The pre-save **Test** validated a file source by *counting* it — a full `list_files` enumeration — and the OneDrive **OAuth callback built the signing-in user's catalog inline on the redirect**, so the browser sat on the callback URL for the length of a recursive walk of their entire drive (OneDrive walks recursively by default; SharePoint does not). Both walks were serial — one Graph `children` round-trip per folder — over a connection rebuilt (TCP + TLS) for every request, because the client used the module-level `httpx.get`. The sign-in walk was wrapped in a best-effort `try/except`, so a Graph throttle midway meant minutes of waiting followed by an empty catalog and a log line.

### Bounded where the cost was unnecessary

- `GraphDriveClient.test_connection` never enumerates: token → site/drive/root → **one 5-item page** of the scoped root, which is what proves access. Every result carries per-step `timings` and a `details` dict, so a slow or failing test says *which* step was slow instead of just spinning.
- `_acount_files_for_validation` takes a `limit` (200 for connection tests), passed only to clients whose `list_files` accepts it. A count that hits the cap is reported as a floor — "Found 200+ files … the full catalog is indexed in the background after saving" — rather than implying the test counted everything.

### Backgrounded where the work is real, with a signal

- The OAuth callback starts a per-user `ConnectionIndexing` row and redirects immediately (`indexing=1` on the redirect). `ConnectionIndexing` gained a nullable `user_id`: NULL is the org-shared run, set is one user's catalog sync. The scopes are queried independently, so a member signing in neither blocks nor is mistaken for the shared run, and one member's sync is never reported as another's.
- The signal follows the work: the runner reports the folders it is enumerating and the files it is indexing; `GET /connections/{id}/indexing` takes `?scope=auto|user|org` (auto prefers the caller's own run); the connection list inlines the viewer's run so the existing card + progress bar light up; and the tables step shows "Building your files list…" and reloads itself when the sync lands.

### Made the walk itself cheaper

- One pooled `httpx.Client` per client instance (keep-alive instead of a handshake per folder), with token acquisition serialized so a concurrent walk doesn't stampede the token endpoint.
- Folders of each level are listed concurrently (`walk_concurrency`, default 8, deterministic output ordering). `walk_concurrency=1` keeps the old depth-first walk.
- `index_mode` (`none` skips the catalog entirely — the tool layer lists and reads live) and `max_catalog_objects`, the same knobs Files & Directories and S3 already had. Outlook Mail keeps its own empty config schema so the drive knobs don't appear on the mail form.

### Behavior changes worth knowing

- **`max_catalog_objects` defaults to 5000** (matching S3), so an existing library/drive with more files is truncated with a warning where it previously indexed everything. It's configurable per connection, and `index_mode: none` skips cataloging altogether.
- **Sign-in no longer returns with the catalog already populated.** The tables step handles this (it follows the job and reloads); other landing pages already poll while indexing is active.
- `POST /connections/{id}/my-schema/refresh` is deliberately left synchronous — it's an explicit user action with a spinner, and it benefits from the pooled/concurrent walk.

### Testing

- New `backend/tests/unit/test_graph_drive_catalog_cost.py` (20 tests): a test never walks; each folder listed exactly once; siblings genuinely overlap (a `threading.Barrier` deadlocks if the walk is serial); `limit` stops early; `index_mode=none` enumerates nothing; the catalog cap holds; progress is emitted for both phases; cancellation unwinds the walk; the bounded count and its "N+" wording; and the knobs surviving registry → client construction.
- New `backend/tests/e2e/test_connection_indexing_user_scope.py` (3 tests): scope isolation across `auto` / `user` / `org`, another member's run not visible as yours, and an in-flight per-user sync not making an admin `/reindex` a no-op.
- New e2e case in `test_connection_oauth_flow.py`: the callback redirects fast with `indexing=1`, the inline sync is **not** awaited, and the per-user row is pollable.
- Updated `test_drive_clients.py` for the bounded probe (plus a case asserting the probe never walks and that failures still carry timings) and `test_schema_validation_file_fast_path.py` for the new `table_count_approximate` field.
- Connection-related e2e suites: 23 passed. Full unit suite: 2096 passed, 16 failed — all 16 reproduce unchanged on `origin/main` (notification/notify service, whatsapp webhook, permissions registry, eval draft helpers, member email, `ms_fabric` OAuth params), verified in a clean worktree.

Alembic: `idxuser01` adds `connection_indexings.user_id` as a plain indexed nullable column (SQLite can't `ALTER` a constraint in, and the column is only ever filtered on, never joined) and leaves a single head.

https://claude.ai/code/session_01DoRnigthQqQnQ5cWjreEy7

---
_Generated by [Claude Code](https://claude.ai/code/session_01DoRnigthQqQnQ5cWjreEy7)_


yochze:
Merged `main` (#788, all-document-libraries SharePoint mode) — it landed in the same file, so a note on how the two combine:

- `list_files` fans out over every library **inside** the bounds: the walk of each library is the new concurrent one, and `limit` is spent across the fan-out, so a bounded listing stops before reaching every library instead of walking N libraries serially. Library-prefixed paths and `drive|item` qualified ids are unchanged.
- `test_connection` keeps the multi-library message ("Indexing 6 document libraries: …") but probes only the **primary** library's root — reaching one already proves the token, the site and the folder scope, and probing each would rebuild the cost this check exists to avoid. `details.library_count` reports the full count.
- `_collect_children` (the shared classify step behind both walks) carries `_qualify_item_id`, so ids round-trip to their own library from either walk.
- Two new cases pin the interaction: the fan-out honoring `limit`, and the test probing one library out of six.

One thing worth flagging: `test_drive_clients.py::test_graph_search_results_carry_relative_path` was already failing on `main` — #788 moved `search_files` onto `_resolve_drives()` while the test still patches `_resolve_drive_id`, so it hit a real `/me/drive` call and `KeyError: 'id'`. Fixed here (patch the new seam). The unit suite isn't in CI (`-m e2e`), which is why it went unnoticed.

Post-merge: 160 graph/file unit tests and 23 connection e2e tests pass locally.

---

PR #810, opened 2026-07-27
Make the All-instructions filters multi-select, and translate the modal

yochze:
Follow-up to #809. Two things: the filters, and the nine locales that PR knowingly left on English.

## The filters

They were native `<select>` elements — so they rendered as OS menus that ignore the app's styling entirely, held one value each, and offered no way to search a list of agents that's already 10+ long in a real org.

Replaced with controls shaped to what each filter actually is:

| Filter | Control | Why |
|---|---|---|
| Agents, People | `MultiSelectFilter` — popover of checkbox rows, search past 8 entries | Unbounded lists that need scanning |
| Made by | `FilterChipGroup` — four chips, always visible | Four options that never grow; a menu costs two clicks and a scan for something that fits in a row |

Plus:

- **Active filters render as removable chips** under the bar. A multi-select trigger collapses to `2 agents` once there's more than one selection, so without the chips the only way to read the current state is to open each menu and count checkmarks. The chips are the readable half of the control.
- **Clearing is one click on the trigger**, not buried in the panel — it's the most common correction.
- **A filter that can't filter is hidden.** The people dropdown doesn't render for an org with one person (the "Anyone / yoyo" menu in the report), and the divider goes with it.

Backend: `agent_ids` / `user_ids` / `sources` take comma-separated lists, matching this API's existing convention (`data_source_ids`, `categories`, `source_types`). Values OR within a facet, AND across facets; blanks are ignored; one bad value in a list is still rejected.

## Two bugs the sandbox caught that tests wouldn't have

**The filter bar rendered as an empty strip.** Components under `components/instructions/` auto-import with a path prefix (`InstructionsMultiSelectFilter`), so the bare `<MultiSelectFilter>` tag resolved to nothing and silently rendered nothing at all. Now imported explicitly, matching how `GitConnectionButton` is already wired in this tree. Worth knowing for anything else added under that directory.

**A stray divider** survived both dropdowns being hidden.

## Localization

#809 shipped the modal wired for i18n but with keys only in `en.json` — and since `plugins/i18n.ts` sets `fallbackLocale: 'en'` with `fallbackWarn: false`, the other nine rendered English *silently*. All **51 keys are now translated across all ten locales**, with placeholder parity asserted programmatically (`{n}`, `{a}`, `{name}` survive every translation).

Verified in French end to end: `Toutes les instructions`, `Historique`, `Tous les agents`, `Effectué par` all render, **zero English leakage**.

## RTL

Translating the strings is the easy half. Running the modal at `locale=he` surfaced three directional bugs that no amount of translation would have caught:

- **The delta counts read backwards.** `+1` rendered as `1+`, `−5` as `5−`. The sign is a bidi-neutral character, so under an RTL paragraph direction it gets reordered to the far side of the digits — inverting the single signal that feed exists to convey. The clock time went the same way (`PM 09:39`). Both now carry `dir="ltr"`, as does the `+`/`~`/`−` op marker on expanded change rows.
- **`text-right`** on the load-mode column pinned it to the wrong edge. Now `text-end`.
- **`pl-[56px]`** on the expanded-row indent indented from the wrong edge. Now `ps-[56px]`.

Verified in the sandbox at `he`: `html dir="rtl"`, zero English leakage, no horizontal overflow, and the filter bar, multi-select popover, active-filter chips and expanded rows all mirror correctly.

## Verification

Sandbox with 10 agents (two Hebrew-named) and 5 instructions, driven through Playwright:

- The agent menu opens with a search box and 10 checkbox rows
- Selecting two produces `agent_ids=<a>,<b>` on the wire
- "Made by" chips render inline, no menu
- The people filter is correctly absent (one person in the org)

Tests: 1 new (multi-value OR-within/AND-across, whitespace tolerance, one-bad-value rejection). **40 passed** across `test_instruction_activity`, `test_instruction_light_view`, `test_instruction`. SQLite only — no Docker for Postgres here.

## Not in this PR

The hunk-level changelog detail we discussed. Accepting a hunk already creates a build so it appears; **rejecting one doesn't** — `reject_hunk` only writes `rejected_hunks` and leaves the snapshot immutable, so that decision leaves no trace a build-derived feed can see. That needs its own event source. And rendering hunks on expand would pull in the O(n²) `difflib` path, which is still unfixed.

https://claude.ai/code/session_01HmKSxiniJKW8o2WRPFAK85

---
_Generated by [Claude Code](https://claude.ai/code/session_01HmKSxiniJKW8o2WRPFAK85)_


yochze:
Pushed a follow-up (`ad982d3`) after verifying the modal in Hebrew — the body above only covered French, and RTL turned up three real bugs that an LTR locale can't show.

**The delta counts were rendering backwards.** `+1` came out as `1+`, `−5` as `5−`. The sign is a bidi-neutral character, so under an RTL paragraph direction it gets reordered to the other side of the digits — which inverts the one signal that feed exists to convey. A reader scanning the changelog in Hebrew would have read every addition as a removal-shaped token and vice versa. The clock time reordered the same way (`PM 09:39`). Both are now `dir="ltr"`, as is the `+`/`~`/`−` op marker on expanded change rows.

Two layout fixes alongside it: the load-mode column used `text-right` (pinned to the wrong edge in RTL) and the expanded-row indent used `pl-[56px]` (indented from the wrong edge). Now `text-end` and `ps-[56px]`.

**Verified in the sandbox at `locale=he`:** `html dir=rtl`, zero English leakage across the list and changelog tabs, no horizontal overflow, and the filter bar, multi-select popover (checkbox on the start edge, Hebrew search placeholder) and active-filter chips all mirror correctly. The trigger pluralizes properly too — `2 סוכנים`.

Worth noting for anything else added to this tree: the logical-property classes (`ps-`/`pe-`/`text-start`/`text-end`) handle mirroring, but they do nothing for bidi reordering *inside* a run of text. Signed numbers and times need an explicit `dir="ltr"` regardless.

---

PR #846, opened 2026-08-01
Scheduled task titles + Automations UX; live report status indicators + real-time sort; bump 0.0.505

yochze:
## What

Scheduled tasks had no name of their own — lists, notifications, and spawned run-reports all borrowed the host report&#39;s title, which breaks down with multiple schedules per report and reads as &#34;Untitled Report&#34; in notifications. This PR gives scheduled tasks a real title end-to-end and brings the Automations lists (Scheduled + Triggers tabs) to the same UX.

### Backend
- New nullable `scheduled_prompts.title` column (migration `sptitle01`, based on `297905a87c8a`). No backfill: every read site falls back `title → report title → prompt snippet`, so existing rows are unchanged.
- Create/update schemas + service accept `title` (explicit empty string clears it); run notifications say `&#34;&#34; ran` and spawned run-reports are named ` — `.
- `GET /scheduled-prompts` gains `status=all|active|paused`; search now matches task titles as well as report titles.
- AI tools: `create_scheduled_task` now **requires** a short title (the agent uses the user&#39;s wording or generates a 2–5 word name); `edit_scheduled_task` can rename. The `` context block and the conversation-history tool digests include the title so follow-up turns can reference tasks by name.

### Frontend
- **Scheduled tab**: title as the card headline (schedule + last run below), per-row active/paused toggle (optimistic, reverts on failure), and All | Active | Paused filter tabs.
- **Triggers tab**: matched to the same layout — trigger name as headline, search, All | Active | Paused tabs (client-side; the owner-scoped list is unpaginated), and the static badge replaced by the same pause/resume toggle via partial `PUT /triggers/{id}`.
- Schedule modal gains a title field (create + edit); shared task card and in-chat tool cards show titles.
- New i18n keys in all 10 locales.

## Also included — live report status indicators (merged from #848)

Merged at 86a89be per maintainer request; full design/evidence write-up in #848's description.

- **Per-report live status badges** on every list surface (sidebar, /reports, project pages, home cards): spinner while running, pulsing amber when the run is **waiting for the user** (clarify form / tool approval), amber for errored-since-last-look, blue dot for unread activity, nothing when read + idle. Precedence: awaiting → running/queued → error+unread → unread.
- **Real-time re-sort**: reports jump to the top of lists the moment new activity lands, no refresh.
- **New pipeline**: per-user `report_views` watermark (migration `rptview01`), `GET /reports/activity` snapshot, `POST /reports/{id}/viewed`, and `GET /reports/activity/stream` — an authenticated per-user SSE feed backed by a per-worker DB watcher (~2s tick, runs only while subscribed, diffs and fans out changes; correct across uvicorn workers because the DB is the bus).
- **Removes the old report WebSocket stack** — it was unauthenticated and dropped events for clients on other workers; the agent's same-worker steering fast path survives as `app/streaming/completion_event_bus.py`. Report-page cross-tab refresh now rides the activity stream.
- **Scale-validated** (harness in `backend/tests/load/`): 100 users / 100 concurrent SSE streams / 500 reports flipping state in one transaction, LLM mocked as direct DB transitions — 0 dropped events across 100k deliveries, single-change delivery ~1.1s to all clients, mass-burst p95 3.7s.
- Alembic: no-op merge revision `mrgheads04` joins the `rptview01` and `whproj01` heads.

## Release
- Bump to **0.0.505** with a combined changelog entry for both features.

## Tests
- `tests/unit/test_scheduled_task_tool.py`, `tests/unit/test_scheduled_prompt_cron.py` — 46 passed (updated for the required tool title + digest).
- `tests/e2e/test_scheduled_prompt.py` — 14 passed (title optional at the API layer, existing flows unchanged).
- `tests/e2e/test_report_activity.py` — 8 passed (state derivation, per-user unread, clarify pause, watcher tick emit/dedup, org scoping, stream auth).
- Merged tree verified: report-activity + scheduled-prompt + queue/steer suites green (26 passed).

https://claude.ai/code/session_01H25niwxrRCDkvyX1WZeo8Q


yochze:
### CI: `e2e-tests (sqlite)` failure was pre-existing — fixed here

`test_daily_usage_series_groups_events_by_day_and_zero_fills` failed with `assert 142 == 42`. Not caused by this PR: it reproduces on a clean `origin/main` worktree (`f6d40f1`).

**Root cause — it only fails on the 1st of the month** (today). The test seeds an "earlier day" event of 100 tokens at:

```python
earlier = max(window_start.date(), today - td(days=3))
```

When the month window starts today, `earlier` collapses onto `today`, so those 100 tokens land in today's bucket and it totals 142 instead of 42. The matching *assertion* is already guarded with `if earlier != today:` — only the seeding wasn't, so the event still got written. Fixed by guarding the seed the same way; the test now passes on any calendar date (18/18 in the file).

### Also pushed: trigger setup redesigned around the delivery URL

Follow-up work in the same automations area, verified end-to-end in a local sandbox (real backend + Nuxt + live Haiku run):

- **Secret-URL is now the default auth mode.** The previous default (token header) 401s every provider that can only be handed a link — Intercom, Zapier, cron jobs.
- **The delivery URL is visible from the moment the modal opens.** "New trigger" provisions an inactive trigger up front, so there's no create-vs-edit mode and no duplicate-create bug.
- **An inactive trigger records deliveries instead of 404ing**, so setup can confirm the event actually arrived and show the payload to write the task against. A 4xx there also makes senders auto-disable the endpoint, which would leave a paused trigger silently dead after resume.
- **GET/HEAD on the delivery URL return 200** without a DB hit — providers probe the endpoint before accepting it, and answering differently for real vs bogus tokens would leak which tokens exist.
- **Rotation in secret-URL mode mints a new path token.** Rotating only the signing secret left a leaked URL fully working.
- Inbound webhook events in a session expand to show the delivery (body + headers, credential-bearing headers stripped).

Plus one more latent bug: `create_scheduled_prompt` pinned `is_active=True` while the create schema advertised the field, and registered the cron job regardless — a task created paused would have fired.

Local: 94 passed across `test_triggers`, `test_scheduled_prompt`, `test_scheduled_spawn_routing`, `test_scheduled_task_tool`, `test_scheduled_prompt_cron`, `test_usage_limits`.

---

yochze:
### `integration-data-sources` failure is a Docker Hub flake, not the diff

Both failures on run [30696765977](https://github.com/bagofwords1/bagofwords/actions/runs/30696765977/job/91360864192) are testcontainers failing to pull its reaper image:

```
docker.errors.APIError: 500 Server Error for
http+docker://localhost/v1.48/images/create?tag=0.8.1&fromImage=testcontainers%2Fryuk:
Internal Server Error ("Get "https://registry-1.docker.io/v2/": context deadline exceeded")
```

`test_data_source_connection[postgresql]` and `[mysql]` never got a container; the 2 tests that don't need one passed. The commit under test (`3485825`) changes a Vue prop and 10 locale JSON files — nothing that can reach Docker or a data-source client.

Needs a re-run of that job; I don't have permission to trigger one from here (`rerun-failed-jobs` → 403). `integration-llms` is green on the same commit; `e2e-tests` (sqlite + postgres) and `playwright-tests` are still running and I'll report back when they land.

---

yochze:
### `e2e-tests (sqlite)` failure: clock-dependent test, fixed

`test_refresh_on_view_single_flights_once_the_gate_opens` failed with `assert False is True` — not caused by this PR's changes, but worth fixing rather than re-running, because it will recur.

The test takes two views and expects the second to be turned away by the claim the first spent. The claim key is a wall-clock bucket:

```python
bucket = int(time.time() // window_seconds * window_seconds)   # window = REFRESH_ON_VIEW_MIN_INTERVAL_SECONDS = 300
```

so both views have to land inside the same 5-minute bucket. Start near a bucket edge and the second view gets a *fresh* claim, wins it, genuinely reruns, and `skipped` is legitimately `False`. The failure landed at 13:47 on a heavily-loaded runner, which is exactly when the gap between the two calls is widest.

Fixed by parking until the current window has room before the first view (waits only when <10s remain, so ~3% of runs pay a few seconds). Same class as the daily-usage failure earlier in this PR: the assertion was correct, the setup was clock-dependent.

Locally: 7/7 in `test_report_refresh_on_view.py`.

Note this failure was reported against `5b2a7769`, which is now four commits behind — `033e3179` carries the fix and a fresh run is underway.

---

PR #850, opened 2026-08-01
feat(mcp): reference MCP tools from instructions; govern native registration per org

yochze:
Instructions can now be linked to an MCP tool the same way they are linked to a table, and native MCP tool registration moves from a deployment env var to a per-org setting that engages adaptively on catalog size.

Both existed as *intent* already — the `connection_tool` reference type has been in the model since #769, and native registration since #797 — but the first was unreachable from either instruction-authoring surface, and the second could only be switched on by editing deployment environment variables for the whole instance.

_No linked issue — raised in conversation while auditing per-agent MCP tool controls._

---

## Executive summary

**For a customer running a large MCP catalog (e.g. an ERP connector with 40+ tools), two things were broken.**

1. **They could not teach the agent how to use a tool.** Knowledge about a tool — "always pass the company code", "this endpoint needs a date range" — is exactly what instructions exist to capture. The product advertises tool references (the mention picker offers them, the data model stores them, chips render for them), but on both instruction-authoring surfaces the link either could not be selected at all or was silently discarded on save. A customer would author the guidance, see it accepted, and get an agent that never received it — with no error to explain why.

2. **The efficiency win for large catalogs was locked behind ops.** Native registration measured at half the planner turns and 57% fewer context bytes on a 60-tool catalog. Reaching it required a deployment env change applying to every org on the instance — not something an admin could evaluate, enable for their workspace, or roll back.

---

### 4a — Product/UX impact: linking an instruction to a tool

```mermaid
flowchart TB
    subgraph BEFORE["❌ Before"]
        direction TB
        A1["Admin scopes instruction<br/>to agent with MCP tools"] --> A2["References picker<br/><b>lists tables only</b>"]
        A1 --> A3["Types @get_contract_deliverable<br/>in the text"]
        A2 --> A4["Saves"]
        A3 --> A4
        A4 --> A5(["Reference silently dropped<br/><b>agent never receives the guidance</b>"])
    end
    subgraph AFTER["✅ After"]
        direction TB
        B1["Admin scopes instruction<br/>to agent with MCP tools"] --> B2["Picker lists tables<br/><b>+ the agent's enabled tools</b>"]
        B1 --> B3["Types @get_contract_deliverable<br/>in the text"]
        B2 --> B4["Saves"]
        B3 --> B4
        B4 --> B5(["connection_tool reference persisted<br/>chip shown, survives reload"])
    end
    style A5 fill:#fee,stroke:#c00,color:#000
    style A2 fill:#fee,stroke:#c00,color:#000
    style B5 fill:#efe,stroke:#0a0,color:#000
    style B2 fill:#efe,stroke:#0a0,color:#000
```

### 4b — Technical mechanism

**(i) Why the reference was dropped** — two independent causes, one per surface:

```mermaid
flowchart LR
    subgraph KE["KnowledgeExplorer (agents page)"]
        K1["refOptions built from<br/>agentTables only"] --> K2["every added id hardcoded<br/>object_type='datasource_table'"]
        K3["@mention fires<br/>mention-selected"] --> K4["<b>no handler</b>"]
        K4 --> K5["save body sends<br/>draft.references only"]
        K5 --> K6(["tool never in payload"])
    end
    subgraph GM["Global instruction modal"]
        G1["options fetched<br/><b>once on mount</b>"] --> G2["scope set to agent<br/><i>after</i> mount"]
        G2 --> G3["backend returns tools<br/>only when scoped —<br/>stale list has none"]
        G3 --> G4["validateSelectedReferences<br/>filters against stale list"]
        G4 --> G5(["tool reference deleted"])
    end
    style K6 fill:#fee,stroke:#c00,color:#000
    style G5 fill:#fee,stroke:#c00,color:#000
    style K4 fill:#fee,stroke:#c00,color:#000
```

**(ii) Native registration: one decision, two consumers.** Making the decision depend on a *count* created a new hazard — if the planner and the `<mcp_tools>` renderer computed it independently they could disagree, and a tool's schema would land in both places (wasteful) or in **neither** (stranding the agent on a `search_mcps` round trip):

```mermaid
flowchart TB
    S["org settings<br/>enable_mcp_native_tools<br/>threshold=12 · max=60"] --> R
    E["BOW_MCP_NATIVE_TOOLS<br/><i>overrides both ways</i>"] -.-> R
    R{{"native_tools_enabled(settings, effective_count)"}}
    R -->|"count &ge; 12"| N["planner: one native tool each<br/><b>schema in tools array</b>"]
    R -->|"count &lt; 12"| G["gateway: execute_mcp<br/><b>schema inlined in &lt;mcp_tools&gt;</b>"]
    B["builder resolves ONCE,<br/>report-wide, then stamps<br/>each agent section"] --> R
    B --> RD["renderer reads the stamp,<br/><b>never recomputes</b>"]
    style N fill:#efe,stroke:#0a0,color:#000
    style G fill:#efe,stroke:#0a0,color:#000
    style RD fill:#eef,stroke:#00a,color:#000
```

---

## Reviewer decision box

| | |
|---|---|
| **Risk** | Medium — one intended behavior change ships on by default (see below); the instruction-reference fixes are additive. |
| **Blast radius** | Instruction authoring on 2 surfaces; MCP tool registration + `<mcp_tools>` rendering for **every org** with ≥12 effective MCP/custom-API tools. Orgs below the threshold are unaffected. |
| **Behavior change (intended)** | **Yes.** Native MCP registration previously required `BOW_MCP_NATIVE_TOOLS=1`; it now defaults **on** for reports with ≥12 effective tools. Backed by the benchmark in #797 (level at 12 tools, 2× better at 60). Admins can switch it off per-org; the env var still overrides in both directions. |
| **Verified ✅** | 44 backend unit tests observed passing post-merge (registry, schema placement, reference scoping), incl. defaults read through the real `OrganizationSettings`. `KnowledgeExplorer.vue` compiles (`@vue/compiler-sfc`). No DB migration needed — defaults resolve from the schema when the key is absent (pinned by test). |
| **NOT verified ⚠️** | No live UI/e2e run of either instruction surface. No live-LLM validation of native mode after this refactor. Full `tests/unit/` sweep was started but **interrupted before completion — not observed**; only the targeted suites below were. Frontend has no test runner exercise beyond the compile check. |
| **Where to look first** | `mcp_tool_registry.native_tools_enabled()` and `schema_context_builder._apply_native_mcp_decision()` — these two **must** agree for the same report, or schemas land twice or nowhere. Then the default in `organization_settings_schema.py` (this is the line that flips behavior for existing orgs). |

---

## Evidence / KPI

| Metric | Result |
|---|---|
| Merge with `origin/main` (23692f9) | Clean, no conflicts |
| `test_mcp_tool_registry.py` + `test_mcp_native_schema_placement.py` + `test_instruction_tool_references.py` | **44 passed**, 0 failed — observed post-merge |
| Adjacent context/render suites (powerbi, agent-status, source-metadata, file-source, inactive-tables, ds-dropped, execute-mcp-routing) | **40 passed**, 0 failed — observed pre-merge, after the final render fix |
| `KnowledgeExplorer.vue` SFC parse + script + template compile | Pass |
| Full `tests/unit/` sweep | ⚠️ **Not observed** — run interrupted |
| Failures attributable to this PR | **0 observed.** No pre-existing noise to attribute away in the suites that were run — every one was green, so no attribution appendix is needed. |
| New tests added | 16 (8 settings/threshold resolution, 2 real-`OrganizationSettings` defaults, 5 schema-placement contract, 1 backend reference scoping) |

One regression was **self-introduced and caught by the new tests**: replacing the renderer's local `native_on` left a dangling reference on the `elif` branch (`NameError` on every `<mcp_tools>` render with a non-inlined catalog). `test_mcp_native_schema_placement.py` failed on it before it reached the commit — it is fixed in 054deff.

---

## Context map

**Key files**

| Path | Role |
|---|---|
| `backend/app/ai/tools/mcp_tool_registry.py` | The decision primitive: `native_tools_enabled(settings, count)`, budget, threshold, env override |
| `backend/app/ai/context/builders/schema_context_builder.py` | `_apply_native_mcp_decision()` — resolves report-wide, stamps each section |
| `backend/app/ai/context/sections/tables_schema_section.py` | Renderer; reads the `native_mcp` stamp, never recomputes |
| `backend/app/schemas/organization_settings_schema.py` | The 3 new lab settings + their defaults |
| `frontend/components/KnowledgeExplorer.vue` | Agents-page authoring: `refOptions`, `onEditorMention` |
| `frontend/components/InstructionGlobalCreateComponent.vue` | Global modal: re-fetch options on scope change |

**Convention this PR establishes**

- A tunable that changes agent behavior belongs in `OrganizationSettings` as a `FeatureConfig` (`is_lab=True` while it soaks), **not** an env var. Env stays as a deployment escape hatch that overrides in both directions.
- When a decision is consumed by more than one component, resolve it **once** at the level that owns the scope and pass the result down. Do not let two components recompute the same predicate from different inputs.
- Reference-type handling (`object_type`) must be carried from the option that produced it — never hardcoded at the insertion site.

**Commits by role**

| SHA | Role |
|---|---|
| `9e12029` | Fix — global modal: re-fetch reference options when scope changes |
| `5859ba4` | Fix + feature — KnowledgeExplorer: tools in the picker, mentions captured |
| `054deff` | Feature + refactor — org settings, adaptive threshold, single-source decision |
| `b53126a` | Merge `origin/main` |

**Follow-ups (deliberately out of scope)**

- **Tool-linked instructions are stored and displayed, but not yet used at runtime.** Table references get retrieval boosting and a `tables="…"` attribute in context; `connection_tool` references get neither. Wiring them in — surfacing an instruction next to its tool's schema at planning time, and/or on the tool result — is the natural next step and the reason most people want the link.
- Native registration past `mcp_native_tools_max` leaves the remainder on the gateway with no inline schema. Harmless today (the inline cap is 40 and the budget is 60, so the overflow regime never inlined anyway), but it is a real edge if either number is retuned.
- The adaptive threshold (12) is taken from the #797 benchmark's "level at 12, 2× at 60" datapoints. It is a reasonable interpolation, not a measured optimum.

---

### Root cause — instruction → tool references

Two unrelated defects, one per surface, both silent:

- **KnowledgeExplorer** (`/agents`): `refOptions` was assembled purely from `agentTables`, and `refIds`'s setter stamped `object_type: 'datasource_table'` on every id it received. There was no path by which a tool could enter the list. Separately, the editor's `mention-selected` event had **no handler here**, while `save` sends only `draft.references` — so @-mentioning anything, tool *or table*, produced no reference row on this surface at all.
- **Global modal**: options are fetched once in `onMounted`. The backend returns `connection_tool` rows only when the request carries a data-source scope (`instruction_service.py:4053`, deliberate — tools resolve per agent through the overlay). Scoping the instruction *after* the modal opened left a tool-less list, and the `selectedDataSources` watcher then ran `validateSelectedReferences()` against it, deleting any tool reference the user had added by mention.

The backend was correct throughout — `test_instruction_tool_references.py` calls `get_available_references` scoped to an agent with MCP tools and gets them back. That test is kept as the regression guard.

### Approach — native registration

Roads not taken:

- **Flip the env var default to on.** Simplest, but leaves the control at instance granularity and gives small-catalog orgs a new code path for no measured benefit.
- **A string-enum setting (`auto`/`on`/`off`).** The settings UI renders booleans as toggles and numbers as inputs; a string falls through to a free-text box. A bool + an integer threshold expresses the same three modes with affordances the UI already has.
- **Thread the actually-registered tool names into the renderer** so it could inline schemas for exactly the budget overflow. Strictly more correct, but requires plumbing planner state into context building; the current cap arithmetic makes the gap unreachable, so it is recorded as a follow-up rather than built.

The threshold is measured on **effective** tools — enabled for the agent via the `DataSourceConnectionTool` overlay and not policy-denied for the requesting user — not on everything the servers advertise, so a catalog that is mostly disabled does not drag the report over the line.

### Scope

Tool *policy* and *enablement* semantics are untouched: the org → per-agent overlay → per-user preference stack and `allow`/`ask`/`deny`/`auto` resolve exactly as before, and native calls are still rewritten onto the `execute_mcp` path so identity forwarding, materialization and audit stay on one code path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PMKSRWZEkFMoK3w8zYUZFN)_


yochze:
## CI: `e2e-tests (sqlite)` was red — pre-existing, fixed in 4175e17

`test_daily_usage_series_groups_events_by_day_and_zero_fills` failed with `assert 42 == 142`. **Not caused by this PR** — the file is untouched by the rest of the branch (`git diff origin/main...HEAD -- backend/tests/e2e/test_usage_limits.py` is empty). It fails on `main` and on every open branch, but only on the **1st of the month**, which today is.

### Root cause: two fixes for the same flake landed today and contradict each other

| Commit | Time (UTC) | What it changed | Effect on the 1st |
|---|---|---|---|
| `05f08f2` | 10:14 | **Seeding** — skips the 100-token event when `earlier` collapses onto today | today's bucket holds **42** |
| `b257e64` | 14:33 | **Assertion** — expects `142 if earlier == today` | test demands **142** |

Each is self-consistent on its own. Composed, the test seeds 42 and asserts 142. `b257e64`'s message notes it was "surfaced on this PR's CI run dated Aug 1" — the same calendar trigger, fixed from the other end, without the seeding guard being visible yet.

### Fix

Kept `05f08f2`'s seeding guard — it also guards the `if earlier != today` assertions on the earlier-day bucket just below, so it's the half that describes the whole test — and dropped the collapsed-total branch from the assertion.

### Evidence (run locally on 2026-08-01)

```
without the change:  E  assert 42 == 142      1 failed
with the change:                              1 passed
```

Reproduced by stashing only this hunk, so the before/after differ in exactly that expression.

### Note for reviewers

This is an unrelated test-only fix riding along in a feature PR, which I'd normally avoid. Including it because the failure blocks this PR and every other PR opened today, and it disappears by itself tomorrow — which would leave the contradiction in the tree to resurface on 1 September. Happy to split it into its own PR if you'd rather land it separately and rebase this one.

---

PR #871, opened 2026-08-02
Stop cutting content out of exported report PDFs

yochze:
A scheduled report emailed as PDF arrives with content missing: the right-hand columns of a wide table gone, long tables stopping after the rows that happened to be on screen, charts sliced at the page edge.

## Why it happened

The renderer treated a PDF like a screen.

A dashboard is built to scroll — the codegen prompt tells the model to wrap wide tables in `overflow-x-auto` (`create_artifact.py`), and the built-in `TableCard` caps its height with `overflow-auto` (`artifact-globals.js`). Paper does not scroll, so everything outside those boxes was silently discarded.

On top of that the document was laid out at a 1280px viewport and then printed to A4 portrait, re-flowing into ~700px. ECharts bakes its container size into the canvas at init and `page.pdf()` never fires a resize, so charts kept their desktop width and hung off the sheet.

## What changed

**Render for paper** (`report_pdf_service.py`)

- Expand containers that are actually hiding content before printing. Only boxes that *both* clip and have content beyond the clip are touched, so decorative `overflow:hidden` and ellipsis truncation are left alone. A `max-width` the content has outgrown is relaxed too — that does not clip, it spills the table out of its card and off the page.
- Lay out at the real printable width, on a landscape sheet, and re-size the charts (animations off) after every layout change.
- Fit a dashboard to one page when it can be. Chromium prints at `printable_width / scale`, so choosing a scale is choosing a layout width; solve for the width where the content's aspect ratio matches the sheet. Past `MIN_SCALE` a single page stops being legible, so fall back to full-size pagination with cards, charts and rows kept whole and table headers repeated.
- Measure the furthest painted edge rather than trusting root `scrollWidth`, which under-reports content escaping a centred, width-capped wrapper — exactly where the last column was being lost.
- Add a gutter when content forces the layout wider: a block does not apply its padding to overflowing content, so a table given exactly its own width ends up flush against the page margin and reads as cut.

**Two adjacent exports were broken outright**

- *Slides.* A slides artifact's `content['code']` is python-pptx source, not HTML; feeding it to the HTML renderer emitted three pages of Python. The deck is the artifact, so convert its `.pptx` through the existing LibreOffice helper (`_office_convert.py`). Report-level export now prefers the dashboard and only falls back to a deck.
- *Docs.* Docs are markdown rendered client-side by `DocViewer.vue` and have no `content['code']`, so the HTML path produced a blank page. Refuse instead of attaching an empty PDF.

**Embedded files** are inlined as data URIs for the render — the headless browser cannot authenticate against `/files/{id}/content`, so every `<BowFile>` in an exported dashboard was a placeholder. The helper moved next to the screenshot path that already used it.

## Verification

Booted the full sandbox (backend + frontend + Claude 4.5 Haiku) and had the model build a dashboard from a 36×14 CSV. It independently produced the exact failing pattern — a 14-column table inside `overflow-x-auto` within a centred `max-w-7xl` wrapper. The scheduled-email path (`generate_for_report`) was then run against that report with the committed code and this branch, and the PDFs compared by text extraction:

| | before | after |
|---|---|---|
| detail columns | **6 / 14** | 14 / 14 |
| detail rows | 36 / 36 | 36 / 36 |
| KPIs / charts | 3 / 2 | 3 / 2 |
| page | A4 portrait, 4pp | A4 landscape, 3pp |
| content vs page edge | **612.8pt on a 596pt page — off the sheet** | 790.1pt, margin at 813.1pt |

DOM audit at print time on the real artifact: no element still clipping content, nothing painted past the document box.

Slides were verified the same way on a model-generated deck: before, three pages of `def generate_slides(...)`; after, five pages of the actual presentation.

## Tests

`backend/tests/e2e/test_report_pdf_completeness.py` — five cases covering un-clipping, single-page fit, data-URI inlining, and the doc/slides refusals. The two browser-backed cases skip when Playwright or the vendored JS libs are absent. PDF text is read back with `pypdfium2`, already a dependency.

Note for anyone running the slides path locally: it needs `libreoffice-impress`, not just `libreoffice-core` — the Dockerfile installs it, a bare dev box may not.

## Known gap

Doc artifacts still have no PDF export. They now fail cleanly rather than attaching a blank page; a server-side markdown + `{{viz:}}` renderer is separate work and is not in this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RQspjKWRpG3ceRi3WvFPjq)_


yochze:
CI note: `unit-tests` came back **cancelled**, and it isn't this branch. The job has `timeout-minutes: 30` and the suite is nowhere near done when the cap hits — dependency install eats ~15 min, and the run was at 32% (`tests/unit/test_file_tools.py`) when the runner cut it.

Same thing on the merge base. On `main` @ `b6a9295`, `unit-tests` was cancelled at exactly 30 minutes (06:35:26 → 07:05:13) while every other job in that run went green and it still dispatched the Docker build.

Not touching it here — raising the cap or splitting the job is a repo-wide call, not part of this change. The other checks are running normally.

---

PR #886, opened 2026-08-03
Decide suggestion containment with a linear scan, not a quadratic diff

yochze:
Instruction saves and opens were taking minutes. Profiling one `GET /api/instructions/{id}` on a sandbox reproducing the reported shape (48 instructions, 8 agents, 12 pending, system-prompt sized bodies): **wall 66.9s, of which 66.7s was a single `difflib` call and 0.025s was SQL.**

## Root cause

`suggestion_merge.covers()` asks whether a pending suggestion is already fully contained in another text. It computed a full character-level alignment and then discarded everything except whether any `delete`/`replace` opcode appeared:

```python
sm = difflib.SequenceMatcher(None, small, big, autojunk=False)
for tag, *_ in sm.get_opcodes():
    if tag in ("delete", "replace"):
        return False
```

That is quadratic, and `autojunk=False` disables the popular-element heuristic that keeps difflib usable on repetitive input — which a system-prompt instruction (many near-identical rule lines) is:

| instruction size | `covers()` |
|---|---|
| 2.5k chars | 0.056 s |
| 5k | 0.250 s |
| 10k | 0.831 s |
| 20k | 3.2 s |
| 40k | 15.1 s |
| the customer's real 20k pair | **66.7 s** |

## Why the earlier fix didn't cover it

`docs/feedback-loops/agents-pending-reconciliation-perf.md` (PR #813) capped the sibling diff in `text_hunks` at `MAX_DIFF_TOKENS = 2000` and moved it off the event loop behind `_HUNK_CPU`. Neither guard reaches this one:

| | `text_hunks` rebase | `suggestion_merge.covers()` |
|---|---|---|
| size cap | `MAX_DIFF_TOKENS` | none |
| unit compared | word tokens (~5k for a 20k-char doc) | raw characters (20k) — ~16× the work |
| off the event loop | `asyncio.to_thread` + semaphore | no |

And it sits on the **detail** path — where that fix deliberately routed the authoritative check ("resolves the moment the instruction is opened") — plus the review panel and the response of every `PUT`.

## The change

"Every character preserved, in order, with only insertions between" is exactly *"`small` is a subsequence of `big`"*, which one left-to-right scan decides in O(n+m).

The scan answers the docstring's question **exactly**; difflib only approximated it. Its greedy longest-match can emit a `delete` even where an embedding exists, so it could answer False where the contract says True — never the reverse, since an alignment with no delete/replace leaves every character of `small` in an `equal` block, which *is* a subsequence. 1,000 randomised edit pairs (insert / append / delete / replace / line-reorder / identical) produced no disagreement in either direction.

`superseded_by_containment` calls it O(candidates²) times; the length test added there prunes pairs that cannot match to O(1), since `b` must be strictly longer than `a` to add anything.

## Measured

Same sandbox, same data, before → after:

| request | before | after |
|---|---|---|
| `GET /instructions/{id}` | 68.1 s | **0.05 s** |
| `GET /instructions/{id}/versions` | 3.3 s | **0.05 s** |
| `GET /instructions/{id}/resolved-evals` | 2.9 s | **0.05 s** |
| `GET /instructions/{id}/pending-builds` | 204.8 s | **0.09 s** |
| `PUT /instructions/{id}` (save) | 70.5 s / 133.7 s | **0.19 s** |

Because it ran on the event loop it also starved the worker — the same mechanism the earlier PR described. An unrelated instruction-list query, unchanged:

```
idle                                    0.076 s
alongside one detail request  before   55.7 s
                              after     0.13 s
```

Results are unchanged: the same 12 pending ids out of 48 instructions, and the same suggestion returned by the review panel.

## Tests

- `backend/tests/unit/test_suggestion_merge.py` (new, 17 tests): the contract cases, a differential check against the previous difflib implementation over randomised edits, the one-way-implication guarantee the swap rests on, a perf guard asserting the pathological 20k-char pair stays under 1 s, and the `superseded_by_containment` behaviours including the deletion-only suggestion that must never be dropped.
- Existing suites: `test_instruction.py`, `test_build.py`, `test_instruction_light_view.py`, `test_instruction_resolve.py`, `test_instruction_edit_anchors.py`, `test_search_instructions_pending.py`, `test_instruction_main_build_integrity.py` — 92 passed.

## Not included

Three separate findings from the same investigation, left for their own changes:

1. Every save creates a build snapshotting **every** instruction in the org (`BuildService._copy_build_contents`) — ~1.5 s and 4,000 inserted rows per save at 4k instructions, and `build_contents` grows unboundedly.
2. `AgentReliabilityService._run_build_fanout` derives `has_global` from the whole snapshot rather than the changed rows, so essentially every save fans out to every agent in the org (reproduced: editing one DS-scoped instruction recorded 6 `global_change` runs).
3. Concurrent saves can truncate the main build — `get_or_create_draft_build` can hand out a build another request is still copying into, the collision is swallowed by `create_build`'s `except Exception`, and in a 5-way burst the sandbox's main build went from 4,000 instructions to 2.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WBz5PkzdbnUMC5A9W9wWxE)_

### DIFF
--- a/backend/app/services/suggestion_merge.py
+++ b/backend/app/services/suggestion_merge.py
@@ -13,7 +13,6 @@ We instead detect the relationship by *content*: if one suggestion's text is the
 other's text plus pure insertions, the smaller is an intermediate the larger
 already covers, and only the larger (cumulative) suggestion should surface.
 """
-import difflib
 from typing import List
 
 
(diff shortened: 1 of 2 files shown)


yochze:
CI red on `playwright-tests`, failing on the base branch too — not from this change.

```
[onboarding] › tests/onboarding/onboarding-wizard.spec.ts:59:3 › step 2: configure OpenAI LLM provider

  Locator:  getByText(/successfully connected/i)
  Expected: visible
  Received: <element(s) not found>
  Timeout:  30000ms   (3 attempts)

1 failed, 61 did not run, 3 passed
```

Identical on main @ `9e63c961` ([run 30816014135](https://github.com/bagofwords1/bagofwords/actions/runs/30816014135/job/91693824123)) — same test, same locator, same line 94, same `1 failed / 61 did not run / 3 passed`. That run's `integration-llms` job failed as well, which points at the LLM credential/upstream path rather than anything in the app: the assertion waits on a live "successfully connected" response from the provider-connection test in the onboarding wizard.

This PR is backend-only — `suggestion_merge.covers()` plus a new unit test. Nothing in it is reachable from the onboarding wizard or from provider connection tests, and the first failing test aborts the run before the other 61 execute, so the suite never gets to any instruction coverage.

Green here so far: `integration-data-sources`, `integration-llms`. Still running: `unit-tests`, `e2e-tests (sqlite)`, `e2e-tests (postgres)` — those are the ones that actually exercise this change, and locally the new 17 unit tests plus 92 existing instruction/build e2e tests pass.

I'll re-check and re-run once the base branch recovers.

---

PR #888, opened 2026-08-03
Replace retired Google model ids and refresh the Anthropic presets

yochze:
Every Google preset in the catalog is unusable for a new API key, which is why the chat returned `google returned an error (404): This model models/X is no longer available` on each model the user picked:

| Old `model_id` | Real status |
|---|---|
| `gemini-3-pro-preview` | Listed under "Previous models — Shut down" |
| `gemini-2.5-pro` | Still in Google's model list, but the API 404s any project with no prior usage |
| `gemini-2.5-flash` | Same |

The 404 classifies as `provider_error` (`backend/app/ai/llm/errors.py:264`), which is both façade-retryable and fallback-eligible — so the retry and the fallback chain both walk to sibling Google models that are equally retired.

## Changes

**`backend/app/models/llm_model.py`** — Google presets replaced with ids verified against Google's model and pricing pages:

- `gemini-3.6-flash` — stable, 1,048,576 ctx, $1.50/$7.50 — now the provider default
- `gemini-3.1-pro-preview` — $2.00/$12.00 (tiered above a 200k prompt), selectable but not a default
- `gemini-3.5-flash-lite` — $0.30/$2.50 — small default

The pro tier has no stable id right now, so the default deliberately sits on a stable flash id rather than a preview one — a preview id is exactly what got shut down.

Anthropic: dropped Claude 4.6 Sonnet, added Claude Opus 5 (`claude-opus-5`, 1M ctx, 128K output, $5/$25, vision). The org default stays on Claude Sonnet 5.

**`backend/app/services/llm_service.py`** — removing ids from the catalog alone fixes nothing for existing installs: `_sync_provider_with_latest_models` skipped rows it couldn't find in the catalog, so dead presets stayed enabled and still held `is_default`. Preset rows whose id has left the catalog are now disabled and surrender their default flags, which lets the existing promotion step move the org onto the new default. Reports pinned to a retired model already fall back to the org default, since `get_default_model_for_report` filters on `is_enabled`.

Tests repointed off `claude-sonnet-4-6` to `claude-opus-5` (same 1M context and vision, so the assertions hold); concurrency probe default updated; changelog entry added.

## Verification

Catalog invariants pass — unique ids, at most one default and one small default per provider, all prices > 0.

111 LLM-related unit tests pass: fallback, error classification, model router, Google tool-schema and message translation, image generation, usage metering. The full `tests/unit` run exceeds the 900s container timeout, so I scoped it to the suites covering this change rather than leaving it half-run.

Sources: [Gemini models](https://ai.google.dev/gemini-api/docs/models), [Gemini pricing](https://ai.google.dev/gemini-api/docs/pricing), [forum report on the 2.5 404](``https://discuss.ai.google.dev/t/gemini-2-5-pro-returns-no-longer-available-to-new-users-contradicts-official-deprecation-date-oct-16-2026/176380).``

https://claude.ai/code/session_01N8KuRKZ4xCh7yJoTKxRNGJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01N8KuRKZ4xCh7yJoTKxRNGJ)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Version 0.0.520 (August 3, 2026)
+- **Gemini works again** — every Google model on offer had been retired by Google, so each question came back with "this model is no longer available" and switching models just produced the same error on the next one; the list is now Gemini 3.6 Flash, Gemini 3.1 Pro and Gemini 3.5 Flash-Lite, and existing workspaces are moved off the dead models automatically
+- **Claude Opus 5 is available**, replacing Claude 4.6 Sonnet in the model list
+
 ## Version 0.0.519 (August 3, 2026)
 - **The AI remembers what it already did** — every tool result is now part of the conversation the model actually sees, so it stops re-reading files it just opened and re-running work it just finished; a file it pulls from a connection stays usable for the whole report instead of going stale after one step, and those background fetches no longer show up as attachments in your message box
 
--- a/backend/app/models/llm_model.py
+++ b/backend/app/models/llm_model.py
@@ -118,6 +118,19 @@ LLM_MODEL_DETAILS = [
         "output_cost_per_million_tokens_usd": 15.00
     },
     {
+        "name": "Claude Opus 5",
+        "model_id": "claude-opus-5",
+        "provider_type": "anthropic",
+        "is_preset": True,
+        "is_enabled": True,
+        "is_default": False,
+        "supports_vision": True,
+        "context_window_tokens": 1000000,
+        "max_output_tokens": 128000,
+        "input_cost_per_million_tokens_usd": 5.00,
+        "output_cost_per_million_tokens_usd": 25.00
+    },
+    {
         "name": "Claude Opus 4.8",
         "model_id": "claude-opus-4-8",
         "provider_type": "anthropic",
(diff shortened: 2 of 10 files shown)


yochze:
CI status — two red checks, both blocked on the same thing, and neither is caused by this diff.

**`integration-llms` — the OpenAI test account is out of credits.** All 8 failures are the `[openai]` parametrizations, failing with:

```
openai.RateLimitError: Error code: 429 - {'error': {'message': 'You have no credits remaining.
Add credits to continue using the API ...', 'type': 'insufficient_quota',
'code': 'credit_balance_exhausted'}}
```

Result was `8 failed, 6 passed, 12 skipped, 19 deselected` — every failure `[openai]`, and the Anthropic cases passed, so the catalog edit itself round-trips fine against a live provider.

**`playwright-tests` — same root cause, and already red on `main`.** The failing test is `onboarding-wizard.spec.ts:94` › "step 2: configure OpenAI LLM provider", waiting on `getByText(/successfully connected/i)`. The wizard's connection test makes a real OpenAI call, which returns the same quota error, so the success banner never renders. Run [30822866849](https://github.com/bagofwords1/bagofwords/actions/runs/30827288259) on `main` at `bee7385` — before this branch — fails on the identical assertion with the identical tally (`1 failed, 61 did not run, 3 passed`).

`integration-data-sources` is green; `unit-tests` and both `e2e-tests` legs were still running at the time of writing.

**Not fixing either**, because there's no code change that would: the blocker is billing on the `OPENAI_API_KEY_TEST` account. Adding credits should clear both. I deliberately did not paper over `integration-llms` by extending the workflow's provider skip list — flagging it instead, per the note in `.agents/skills/add-llm-provider-or-model/SKILL.md`. I'll re-check and report when the base branch recovers.

---

PR #898, opened 2026-08-04
Rebuild the eval run detail page: minimal, collapsed by default, report-style rendering

yochze:
## Summary

Rebuilds `/evals/runs/[id]` as a compact, minimal page (net **−427 lines**):

- **All cases collapsed by default** in one flat list — status icon, case name, passed/total, duration, chevron. No cards, no auto-expand.
- **Report-style conversation** when a case is expanded, in miniature: the prompt as a user bubble, thinking as a clamped gray line (click to expand), tool calls as one-line summaries, and assistant answers rendered through `MarkdownRender` — the raw `Started / Update · status=...` event log is gone.
- Conversation content is rebuilt from `/api/tests/runs/{id}/status` completions (the same data the report page renders) and refreshed on a debounce while SSE events stream in; result/run statuses still update event-by-event, so the page stays live during a run.
- **Checks** render as minimal icon rows with one secondary line (judge reasoning, failure message, or observed value) instead of boxed rule cards.
- Header collapses to title + status pill and a single meta line (suite · trigger · duration · build · passed/total); the compare strip slims to one bordered row; model, data sources, and the report link move to a quiet footer inside each expanded case.

## Test plan

Verified live in a sandbox with Playwright against real agent runs:
- Completed multi-case run: collapsed by default; expanded cases show markdown answers, tool summaries, judge reasoning, compare strip.
- Live run: queued icon → streaming conversation appears via debounced refresh → terminal state with checks and Success pill; Stop button present while in progress.
- No console errors or 4xx responses on the page.

https://claude.ai/code/session_01NzQNWrPst16picgPA6ysC6

---
_Generated by [Claude Code](https://claude.ai/code/session_01NzQNWrPst16picgPA6ysC6)_


yochze:
CI status: `playwright-tests` is red on an external blocker, not this diff.

`onboarding-wizard.spec.ts › step 2: configure OpenAI LLM provider` performs a **live "Test Connection" against OpenAI** using the `OPENAI_API_KEY_TEST` secret. That call currently fails (the same account returns `429 insufficient_quota / credit_balance_exhausted` in-app), so "successfully connected" never appears, the spec fails after 3 attempts, and the remaining 61 Playwright tests are aborted. It fails identically on multiple commits of this branch, including ones whose diff is a single Vue file.

No code change here can fix it — it needs credits added to (or rotation of) the `OPENAI_API_KEY_TEST` account. Once that's done, re-running the workflow should go green. The earlier `e2e-tests` failure (`test_connection_get_tables`) is a separate pre-existing flake: the test asserts discovered tables immediately after `/refresh` returns, but refresh only *starts* a background indexing job — a poll-until-indexed loop in the test would fix it.

---

PR #902, opened 2026-08-04
Prevent folding of blocks carrying user-facing messages

yochze:
## Summary
Blocks that carry user-facing messages (content, assistant notes, or final answers) should never be folded into groupings, as this would hide important information from the user. This change extracts message detection logic into a reusable function and applies it consistently across block grouping logic.

## Key Changes
- **New `blockHasMessage()` function**: Centralizes detection of user-facing message prose across three fields:
  - `block.content` — direct block content
  - `block.plan_decision.final_answer` — final answer text
  - `block.plan_decision.assistant` — planner's assistant message
  
  Only non-empty, non-whitespace text counts as a message.

- **Updated `isGroupableBlock()`**: Now uses `blockHasMessage()` instead of only checking `final_answer`, catching all message-carrying blocks.

- **Updated `computeBlockGroups()`**: Applies the same message check when determining if a tool block can be absorbed into a grouping.

- **Documentation**: Added clarifying comments explaining that blocks with user-facing messages must stay visible and break runs, preventing folding that would hide content.

## Notable Details
- Whitespace-only text (e.g., `'  \n'`) is explicitly not treated as a message, allowing such blocks to still be grouped.
- The change ensures consistent behavior across both the grouping decision logic and the run computation logic.
- Test cases added verify that message-carrying blocks split groupings correctly, including edge cases like handled errors with messages.

https://claude.ai/code/session_01MXsJHxQoA8cJBpkDXjn9Lf

### DIFF
--- a/frontend/composables/useBlockGrouping.ts
+++ b/frontend/composables/useBlockGrouping.ts
@@ -12,6 +12,9 @@
  * - Deliverables (create_data, artifacts, docs), clarify, approvals, the plan
  *   note, errors, and in-flight blocks NEVER fold: content and failures must
  *   stay visible, and the live block is the user's status line.
+ * - A block that carries a user-facing message (content / assistant /
+ *   final_answer) NEVER folds and breaks the run — folding would hide the
+ *   message. Only pure consecutive tool blocks group.
  * - ANY run of two or more consecutive chip-class blocks folds into one
  *   "live ticker" line (single morphing label while running, compact summary
  *   when settled) — one visual grammar for every run, no threshold split.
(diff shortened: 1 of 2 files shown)


yochze:
CI status: everything is green here except **unit-tests**, which is failing on the base branch too — it's not caused by this PR.

Details:
- `unit-tests` was cancelled at exactly its `timeout-minutes: 30` cap with the suite only ~27% through (`tests/unit` runs serially, unlike the xdist-parallelized e2e job). The last completed run on `main` (`bee73852`) shows the identical 30-minute cancellation, so the suite has simply outgrown its timeout.
- While it ran, a few backend test failures were visible in the log (`test_data_source_member_email.py` 1F, `test_eval_draft_helpers.py` 3F) — also unrelated to this frontend-only diff.
- All other checks passed on this PR: e2e (sqlite + postgres), playwright-tests, both integration jobs, and Snyk.

I'm not touching the unit-test timeout or those backend failures in this PR since they're out of scope (this PR only changes the frontend block-grouping composable + its tests). Fixes worth considering separately: run `tests/unit` under `-n 2` like the e2e job, and/or raise the job timeout. I'll re-check when the base branch recovers.

---

PR #910, opened 2026-08-04
Fix Splunk connector on deployments that reject wildcard index searches

yochze:
## Problem

On hardened Splunk deployments (Splunk Cloud guardrails, ES-restricted roles) that reject wildcard index searches, the connector broke down end to end — as reported by a customer:

- Schema indexing "found 4–6 objects" but every table came out named `*::<sourcetype>` with **0 columns**: the wildcard `| tstats count where index=*` catalog was rejected, and the `| metadata` fallback silently hardcoded the index as `*`.
- Every `create_data` / `inspect_data` then failed with *"you must specify a specific index, no global wildcards"*, because field sampling, thin-table descriptions, and the system prompt all emitted `index=*` — and the schema never contained a real index name to fall back on.
- The agent's only remaining move was to ask the user which index holds the data.

## Fix (`splunk_client.py` + one config field)

- **Index enumeration without searching**: `_list_indexes()` uses `GET /services/data/indexes` (pure REST — immune to the wildcard ban). When the fast-path wildcard tstats is rejected, the catalog is rebuilt with an explicit OR-list tstats over the enumerated indexes (chunked by 20): `| tstats count where (index=web OR index=app …) by index, sourcetype`.
- **Never store `*` as an index**: the `| metadata` last resort records the index as unknown (`None`) instead of `*`, and unknown-index tables get honest guidance instead of a `search index=*` instruction.
- **Field sampling never emits `index=*`**: unknown indexes are resolved via a scoped tstats lookup (`_resolve_index`); legacy `*::sourcetype` table names resolve the same way in `get_schema`.
- **Self-correcting errors**: a rejected wildcard search now raises with the known index list appended ("Known indexes: web, app, … — rewrite with an explicit index"), so the agent recovers instead of guessing `main`/`default`.
- **Prompt rewrite**: the system prompt teaches explicit indexes and `(index=a OR index=b)` spans; the `index=*` examples are gone.
- **`indexes` config field** on `SplunkConfig`: optional comma-separated scope, and the escape hatch when even index enumeration is denied.
- Catalog degradation now logs warnings instead of `print`.

## Verification

Reproduced and verified in a live sandbox (real Splunk 9.3 + a guard proxy emulating the customer's wildcard-rejecting deployment + Claude 4.5 Haiku driving the real chat UI) — full evidence in `docs/feedback-loops/splunk-wildcard-index-restriction.md`:

- **Before**: 5 tables named `*::sourcetype`, 0 columns; chat turn = 2 failed `create_data` + 1 failed `inspect_data` (verbatim customer error), ending with the agent asking the user for the index.
- **After** (same guard, same prompt): discovery yields `web::access_combined` (16 cols), `app::log4j` (14), `app::json_app` (15), `security::auth_audit` (14), `metrics::collectd` (12); the chat completes in **one** `create_data` — Haiku wrote `search index=app sourcetype="log4j" level=ERROR | stats count by host` — 12 rows, bar chart, zero blocked searches.
- Unit suite: 20 passed (15 pre-existing + 5 new restricted-deployment cases). `tests/e2e/test_data_source.py` + `test_connection.py`: 11 passed.

## Rollout note

Existing installs hit by this bug have `*::sourcetype` tables stored; re-running schema indexing replaces them with real `index::sourcetype` tables (different names — saved instructions referencing `*::` names should be checked). If a deployment also denies the index-catalog endpoint, the new "Indexes" connection field is the workaround.

https://claude.ai/code/session_017zwCqsddhysJPJaD1TdNY7

---
_Generated by [Claude Code](https://claude.ai/code/session_017zwCqsddhysJPJaD1TdNY7)_

### DIFF
--- a/backend/app/data_sources/clients/splunk_client.py
+++ b/backend/app/data_sources/clients/splunk_client.py
@@ -28,6 +28,8 @@ Auth is Splunk-native:
   - `userpass` → HTTP basic against the management port (older on-prem installs).
 """
 import json
+import logging
+import re
 from typing import Any, Dict, List, Optional
 
 import pandas as pd
@@ -36,6 +38,7 @@ import requests
 from app.data_sources.clients.base import DataSourceClient
 from app.ai.prompt_formatters import Table, TableColumn, ServiceFormatter
 
+logger = logging.getLogger(__name__)
 
 MAX_ROWS = 50_000              # hard cap per execute_query
 DEFAULT_LIMIT = 1_000         # when the query spec omits `limit`
@@ -43,6 +46,11 @@ HTTP_TIMEOUT = 180            # seconds per REST request (searches can be slow)
 SAMPLE_EVENTS = 500          # events sampled per sourcetype for field discovery
 DEFAULT_MAX_SAMPLED = 50     # top-K sourcetypes that get field sampling
 CATALOG_SEARCH = "| tstats count where index=* by index, sourcetype"
+INDEX_CHUNK = 20             # indexes per OR-list tstats when wildcards are blocked
+
+# An SPL wildcard-index reference (`index=*`). Hardened deployments (Splunk
+# Cloud guardrails, ES-restricted roles) reject searches containing it.
+_WILDCARD_INDEX = re.compile(r'index\s*=\s*"?\*', re.IGNORECASE)
 
 
 class SplunkClient(DataSourceClient):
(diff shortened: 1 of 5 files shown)


yochze:
CI status note: `playwright-tests` is red on this PR, but the failure is **pre-existing on `main`** and unrelated to this diff.

- Failing test: `[onboarding] › tests/onboarding/onboarding-wizard.spec.ts:59 › step 2: configure OpenAI LLM provider` — times out waiting for "successfully connected" after testing the **OpenAI** provider (uses `OPENAI_API_KEY_TEST`; possibly an expired/invalid test key).
- The [latest `main` run](https://github.com/bagofwords1/bagofwords/actions/runs/30933126537/job/92072477575) (head `55391fd`, before this PR) fails on the exact same test, same assertion (`onboarding-wizard.spec.ts:94`), same 3-retry timeout.
- This PR touches only the Splunk backend connector, its config schema, its unit tests, and a doc — no frontend/onboarding/OpenAI code.

Checks that exercise this PR's code are green so far: `integration-data-sources` ✅, `integration-llms` ✅ (unit/e2e still running at time of writing). Will re-check once the base branch's playwright failure is fixed.

---

PR #917, opened 2026-08-06
Fix agent loop to properly handle empty planner decisions and step exhaustion

yochze:
## Summary

This PR fixes critical issues in the agent execution loop where empty or invalid planner decisions were not properly rejected, and where reaching the step limit without a terminal decision was incorrectly recorded as success. These bugs caused production runs with 100+ blank planner decisions and zero tool calls to be finalized as successful completions.

## Key Changes

- **Empty response detection in planner_v3**: Added validation to reject streams that produce no semantic output (no text, no tool calls, no reasoning, no web searches). These now return a typed `empty_response` error instead of an empty valid decision, enabling the agent loop's retry mechanism.

- **Raw stop reason preservation**: Added `raw_stop_reason` field to `MessageStopEvent` across all LLM clients (OpenAI, Azure, Anthropic, Bedrock) to preserve provider-specific stop reasons for diagnostics when normalized stop reasons collapse to "other".

- **Stream event forensics**: Planner now tracks event type counts during streaming to preserve evidence of what actually came back when a stream ends with no usable output.

- **Improved decision usability check**: Replaced simple truthiness check with comprehensive validation that examines actions, text content, reasoning, and error state to determine if a decision is actually usable.

- **No-progress retry budget**: Added explicit handling for decisions with no executable action. When the invalid retry count is exhausted, the turn now terminates with an `error` status and `no_progress` error code instead of silently continuing.

- **Step limit exhaustion handling**: Added `planner_steps_exhausted` flag to detect when the outer loop completes without a terminal break. These turns now:
  - Record `completion_errored = True` to prevent auxiliary LLM calls (knowledge harness, scoring, title generation)
  - Set agent execution status to `error` instead of `success`
  - Set completion status to `error` with `planner_step_limit` error code
  - Include diagnostic message about reaching the maximum planning steps

- **Knowledge harness skip on error**: Knowledge harness sub-loop is now skipped when a turn ends in error to prevent spawning additional LLM work on failed completions.

## Implementation Details

- The decision usability check now validates: presence of actions, non-empty assistant message, non-empty final answer, non-empty reasoning message, and absence of errors.
- No-progress errors are properly persisted to the completion record with status updates and SSE events before breaking the loop.
- Step limit exhaustion is detected via the `else` clause of the `range(step_limit)` loop, ensuring it only triggers when the loop completes naturally without a `break`.
- Error status propagation ensures failed turns don't trigger downstream operations like scoring or title generation that assume a valid answer exists.

https://claude.ai/code/session_01RsTEGoDJWZ2hteZvrG5z2D

### DIFF
--- a/backend/app/ai/agent_v2.py
+++ b/backend/app/ai/agent_v2.py
@@ -1771,12 +1771,30 @@ class AgentV2:
                         final_decision = evt.data
                         break
 
-                if not final_decision:
+                # A decision object that exists but carries no content (no
+                # action, no text, no reasoning) — or carries a planner error —
+                # is treated exactly like no decision at all. It used to pass
+                # the truthiness check below and get persisted, leaving a blank
+                # "Planning (unknown)" block on otherwise-finished turns.
+                _h_actions = list(getattr(final_decision, "actions", None) or []) if final_decision else []
+                if final_decision is not None and not _h_actions and getattr(final_decision, "action", None):
+                    _h_actions = [final_decision.action]
+                _h_usable = bool(
+                    final_decision is not None
+                    and getattr(final_decision, "error", None) is None
+                    and (
+                        _h_actions
+                        or (getattr(final_decision, "assistant_message", None) or "").strip()
+                        or (getattr(final_decision, "final_answer", None) or "").strip()
+                        or (getattr(final_decision, "reasoning_message", None) or "").strip()
+                    )
+                )
+                if not _h_usable:
                     # One retry: a single malformed/empty reply must not end
                     # the phase empty-handed (the v2 envelope failure mode).
                     empty_decision_retries += 1
                     if empty_decision_retries > 1:
-                        logger.warning("Knowledge harness: no decision twice in a row; stopping")
+                        logger.warning("Knowledge harness: no usable decision twice in a row; stopping")
                         break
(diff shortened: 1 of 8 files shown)


yochze:
CI status: 5 of 6 checks are green (e2e sqlite/postgres, integration-data-sources, integration-llms, playwright). The `unit-tests` check shows **cancelled**, not failed — the job hit its `timeout-minutes: 30` limit at ~27% suite progress with zero test failures.

This is failing on the base branch too, not caused by this PR: on `main` at the same base commit (`37ae3f5`), `unit-tests` was cancelled the identical way ([run 31030806355](https://github.com/bagofwords1/bagofwords/actions/runs/31030806355/job/92390725851), 17:37→18:06), and every recent `main` run of this workflow shows the same 30-minute cancellation. The unit suite has outgrown the job timeout.

The tests this PR touches pass locally: `tests/unit/test_agent_loop_rescue.py` (16 tests, includes the 4 new empty-stream regression tests) plus the client-streaming / tool-dispatch / error-classification suites — 131 passed. Raising `timeout-minutes` for the `unit-tests` job in `.github/workflows/e2e-tests.yml` (or splitting/parallelizing the suite) is a repo-wide CI decision I'll leave out of this PR unless you want it included; I'll re-check when the base is green again.

---

PR #923, opened 2026-08-07
Fix version history to track live published state correctly

yochze:
## Summary
Fixed the version history panel and version diffs to correctly identify and display the live (published) instruction text, rather than relying on the potentially stale row cache. This ensures accurate version comparisons and proper labeling of unpublished pending versions.

## Key Changes
- **Authoritative live state tracking**: Added `mainText` and `mainVersionId` refs that are populated from the `/api/instructions/{id}/review-hunks` endpoint, which provides the actual published content from the main build
- **Improved current version detection**: Replaced simple index-based checks (`i === 0`) with `isCurrentVersion()` helper that compares against the authoritative `mainVersionId`, since pending suggestions can write versions that appear ahead of the live version in the history list
- **Unpublished version labeling**: Added `isUnpublishedVersion()` helper and UI badge to clearly mark versions that were written by pending/rejected suggestions and never published, distinguishing them from the actual current version
- **Fixed version diff direction**: Corrected `viewVersion()` to compare old version (base) → live text (target), so diffs read intuitively (green = "added since v25", red = "removed since v25"). Previously it used the stale `detail.text` as the target, which on instructions with pending changes wasn't actually diffing against current
- **Proper state cleanup**: Ensured `mainText` and `mainVersionId` are cleared when opening new instructions or creating new ones, preventing stale state from affecting version labeling
- **UI improvements**: Added arrow and "current" label in the version diff header to clarify what's being compared
- **Internationalization**: Added `versionNotPublished` translation key across all supported languages

## Implementation Details
- The live text is now computed via `liveText` which falls back to the row cache only for brand-new instructions that don't yet exist in the main build
- Version history list remains newest-first, but the "current" marker now correctly identifies the published version regardless of its position
- The `/review-hunks` endpoint response now carries both the pending suggestions and the authoritative main build state, eliminating the need to trust the row cache for version comparisons

https://claude.ai/code/session_01LRGRSdphqMh2ewyGyykzNP

### DIFF
--- a/frontend/components/KnowledgeExplorer.vue
+++ b/frontend/components/KnowledgeExplorer.vue
@@ -622,6 +622,12 @@
               <div class="flex items-center gap-2 min-w-0">
                 <span class="w-1.5 h-1.5 rounded-full shrink-0" :class="activeSuggestion?.source === 'ai' ? 'bg-violet-500' : 'bg-blue-500'"></span>
                 <span class="text-xs font-medium text-gray-700 dark:text-gray-300 truncate">{{ diff.title }}</span>
+                <!-- Version compares read old → current, so name both ends:
+                     green is what current has that this version didn't. -->
+                <template v-if="diff.versionId">
+                  <UIcon name="i-heroicons-arrow-right" class="w-3 h-3 text-gray-300 dark:text-gray-600 shrink-0 rtl:rotate-180" />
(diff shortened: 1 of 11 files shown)


yochze:
Superseded by #922 — commit `e4fa326` was merged into `claude/agent-permissions-editing-pq9vp6` (merge commit `e8df5c8`), so the version-history fix ships as part of that PR. Closing this one.

---

PR #935, opened 2026-08-08
Fix instruction display inconsistencies across surfaces

yochze:
## Summary

This PR addresses three critical data mutation and visibility issues in instruction display across the Knowledge Explorer, report Agent panel, and transcript tool cards:

1. **Title uppercasing mutation** — The report panel was rewriting stored instruction titles to uppercase on edit
2. **Reference data loss** — Editing instructions from the report panel was silently rewriting reference display text and relation types
3. **Archived instructions visibility** — Archived instructions appeared in the Knowledge Explorer but were hidden in the report panel

### InstructionGlobalCreateComponent.vue
- **Removed title uppercasing mutation** (L372): The `@input` handler that was calling `.toUpperCase()` on the title field has been removed. The `uppercaseTitle` prop now only affects CSS display, never the stored value.
- **Changed `uppercaseTitle` default from `true` to `false`** (L917): This prevents accidental uppercase transformation of titles when the component is used without explicitly setting the prop.
- **Fixed reference round-trip loss** (L1639–1664): 
  - `display_text` and `relation_type` are now preserved when building the instruction payload instead of being dropped/hardcoded
  - Added logic to `handleReferencesChange` to preserve already-selected references' stored metadata when toggling checkboxes
  - Updated `initReferencesFromInstruction` to carry `display_text` and `relation_type` from stored references through to the form
- **Updated form reset logic** (L1671–1673): Removed the uppercasing of `initialTitle` during form initialization

### ReportAgentPanel.vue
- **Added `include_archived: true`** to the instructions fetch query (L934): This ensures archived instructions are loaded in the report panel, matching the Knowledge Explorer's behavior. Users can filter them out via the status filter if desired.
- **Added clarifying comments** explaining the deliberate difference in `include_global` behavior between the two surfaces

## Implementation Details

The reference preservation fix works by maintaining a Map of already-selected references keyed by ID. When the reference selection changes, instead of blindly rebuilding from the option list (which lacks stored metadata), the code now maps each selected ID back to its original reference object if it was already selected, preserving `display_text` and `relation_type`.

The title uppercasing fix ensures that display-only CSS transformations never reach the stored data model. The prop documentation has been updated to clarify this is display-only behavior.

https://claude.ai/code/session_01M9T5AKqoMSf4BFyJJZNFM9

### DIFF
--- a/docs/instruction-surface-consistency-audit.md
+++ b/docs/instruction-surface-consistency-audit.md
@@ -1,0 +1,414 @@
+# Instruction surfaces — consistency audit
+
+**Status:** investigation only. No code changed.
+**Question:** where do the Knowledge Explorer, the report-session Agent panel, and
+the in-transcript instruction tool cards show *different data* for the *same*
+instruction?
+
+**Answer:** in a lot of places. There is no shared "instruction view" component
+and no shared display contract — each surface picks its own endpoint, its own
+projection, its own label functions, and its own body renderer. The list below
+is grouped by root cause, then by concrete symptom.
+
+---
+
+## 0. The surfaces
+
+| # | Surface | File | Body renderer | Metadata source |
+|---|---------|------|---------------|-----------------|
+| 1 | Knowledge Explorer tree row | `components/KnowledgeExplorer.vue` (`InstrLeaf`, ~L3580) | — (title only) | `GET /instructions?view=light` |
+| 2 | Knowledge Explorer detail pane | `components/KnowledgeExplorer.vue` L647–947 | `InstructionEditor` (TipTap, `editable=false`) | `GET /instructions/{id}` (full) |
+| 3 | Agent panel — Overview → Primary instruction | `components/report/ReportAgentPanel.vue` L114–137 | `InstructionText` (**no** `markdown`) | `GET /data_sources/{id}` → hand-built `primary_instruction` dict |
+| 4 | Agent panel — Instructions list row | `ReportAgentPanel.vue` L333–400 | `InstructionText` (plain, 80 chars) | `GET /instructions?view=light` |
+| 5 | Agent panel — instruction detail/edit | `InstructionGlobalCreateComponent.vue` (view mode L7–355, edit mode L358+) | `InstructionEditor` | `GET /instructions/{id}` (re-fetched internally) |
+| 6 | `create_instruction` tool card | `components/tools/CreateInstructionTool.vue` | `InstructionText` (`markdown`, **no refs**) | `tool_execution.arguments_json` / `result_json` |
+| 7 | `edit_instruction` tool card | `components/tools/EditInstructionTool.vue` | `InstructionText` (`markdown`, no refs) + `TrackedChangesView` | `result_json` + `GET /instructions/{id}` |
+| 8 | `read_instruction` tool card | `components/tools/ReadInstructionTool.vue` | **raw `whitespace-pre-wrap` text** | `result_json` only |
+| 9 | `search_instructions` tool card | `components/tools/SearchInstructionsTool.vue` | `InstructionText` (`markdown`, no refs) | `result_json.instructions[]` |
+| 10 | "N instructions loaded" popover | `pages/reports/[id]/index.vue` L493–518 | — | `message._loaded_instructions` |
+| 11 | Knowledge-harness card | `components/KnowledgeGroup.vue` | `TrackedChangesView` | `result_json` |
+| 12 | Knowledge Explorer → All instructions modal | `instructions/AllInstructionsModal.vue` → `InstructionsTable.vue` | — | `view=light`, `live=true` + `live=false` |
+
+Editors alone: **three** (KE inline pane, `InstructionGlobalCreateComponent`
+inline in the panel, `InstructionModalComponent` → the same component in
+`split-layout` mode, opened from `CreateInstructionTool` and the training modal).
+
+---
+
+## 1. Root cause A — different endpoints ⇒ different fields exist at all
+
+`GET /instructions?view=light` returns `InstructionListItemSchema`
+(`backend/app/schemas/instruction_schema.py` L223). It deliberately drops:
+`text` (→ `preview`, 280 chars), `formatted_content`, `structured_data`,
+`user` (only `user_id`), **`references`**, **`description`**, **`evidence`**,
+**`primary_for`**, `reviewed_by`, `thumbs_up`.
+
+`GET /instructions?view=full` returns `InstructionListSchema` — which *also*
+has **no `references`** and no `evidence` / `kind` is present but `primary_for`
+is.
+
+Only `GET /instructions/{id}` (`InstructionSchema`) has the full set.
+
+### 1.1 Dead bindings in the Agent panel list (rows come from `view=light`)
+
+`ReportAgentPanel.vue`:
+
+- **L382–400 — the entire References block can never render.** `inst.references`
+  is not in `InstructionListItemSchema`, so `v-if="inst.references?.length"` is
+  always false. Dead code.
+- **L375–379 — the author + created-date badge can never render.**
+  `inst.user` is not in the light schema (`user_id` only). So the panel list
+  silently shows no author/date, while the KE detail pane (L923–925) shows both,
+  and `InstructionsTable` (L370) shows the creator name.
+- **L347 — `:references` mapping on the preview `InstructionText` is always
+  `undefined`**, so `@Table` mentions in the preview render as plain text there
+  and as indigo chips everywhere else.
+- **L342 — the title fallback is `(inst.preview ?? inst.text)?.slice(0, 60)`**,
+  i.e. a raw 60-char slice *including newlines*. Everywhere else
+  (`instructionRowLabel`, both tool cards) takes `split('\n')[0]` first. Same
+  instruction, different row label.
+
+### 1.2 Primary instruction is a hand-built dict, not the schema
+
+`backend/app/services/data_source_service.py` L1197–1222 builds
+`primary_instruction` by hand with only:
+`id, text, status, category, source_type, load_mode, title, organization_id, references[]`.
+
+Consequences:
+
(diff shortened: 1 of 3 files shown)


yochze:
## CI status: `e2e-tests (postgres)` red on two pre-existing failures — not from this PR

**Green:** `playwright-tests`, `e2e-tests (sqlite)`, `integration-data-sources`, `integration-llms`.
**Red:** `e2e-tests (postgres)` — 2 failed, 1144 passed.
**Cancelled:** `unit-tests` (fail-fast once the postgres job went red).

This PR changes 3 files — `frontend/components/InstructionGlobalCreateComponent.vue`, `frontend/components/report/ReportAgentPanel.vue`, and a docs markdown. No Python, no models, no migrations. Both failures are in backend e2e tests whose last-touching commits are already on `main`, so they cannot be caused by this diff.

### 1. `test_oidc_group_sync.py::test_soft_deleted_group_still_holds_its_name`

```
asyncpg.exceptions.DataError: invalid input for query argument $9:
datetime.datetime(2026, 8, 8, 9, 23, 8, ...) (can't subtract offset-naive and offset-aware datetimes)
[SQL: INSERT INTO groups (... deleted_at) VALUES (... $9::TIMESTAMP WITHOUT TIME ZONE)]
```

The test constructs `Group(..., deleted_at=_dt.now(_tz.utc))` — a timezone-**aware** datetime — but `BaseSchema.deleted_at` is `Column(DateTime)` with no `timezone=True`, so it maps to `TIMESTAMP WITHOUT TIME ZONE`. asyncpg rejects an aware value for a naive column; SQLite accepts it. That is exactly the observed split: the sqlite e2e job passes and the postgres one fails, deterministically.

The fix is one line in either direction (drop the `tzinfo` in the test, or give the column `timezone=True`), but it belongs with the OIDC group-sync work, not in a frontend-consistency PR — picking one of those two is a call for whoever owns that code, since `timezone=True` would be a schema change.

### 2. `rbac/test_rbac_evals.py::test_case_and_run_listings_narrow_to_one_agent`

```
asyncpg.exceptions.ForeignKeyViolationError: insert or update on table "test_results"
violates foreign key constraint "test_results_head_completion_id_fkey"
DETAIL: Key (head_completion_id)=(68239cde-...) is not present in table "completions".
```

A `test_results` row is inserted referencing a completion that isn't committed/visible in the same transaction — a fixture-ordering or isolation problem, again postgres-strict and sqlite-tolerant.

### Why `main` looks green

Every recent `e2e-tests.yml` run on `main` has conclusion `cancelled` — the concurrency group cancels the in-flight run when the next merge lands, and merges have been frequent. So neither failure has had a completed run on `main` to surface it; they were introduced by work that merged yesterday and went unobserved.

### What I'm doing

Not fixing these here — they're unrelated to this change and each needs a decision from the code's owner. I'll keep watching the PR and re-run against the base branch once these land a fix, so this PR's CI can go green on its own merits.

---

yochze:
## Correction + actual root cause

Two things in my previous comment were wrong. Correcting them, because the real explanation is more useful than the one I guessed at.

### Proof this is pre-existing (stronger than what I claimed before)

Workflow run [31246843832](https://github.com/bagofwords1/bagofwords/actions/runs/31246843832) is a **push to `main`** at head SHA `8457877b9e90b09f6560d9980cb9e1ebdb467952` — which is exactly this PR's base commit. On that run:

| Job | Conclusion |
|---|---|
| `e2e-tests (postgres)` | **failure** |
| `e2e-tests (sqlite)` | success |
| `unit-tests` | cancelled (30m14s) |
| `playwright-tests`, `integration-*` | success |

Same job, same failure, on `main`, before this branch existed. I previously inferred this from commit ancestry; it's now directly observed.

### ❌ Wrong: "`unit-tests` cancelled by fail-fast once postgres went red"

It's a **job timeout**. `unit-tests` has `timeout-minutes: 30`; it started 08:30:43 and was killed 09:00:59 — 30m16s. It died *42 minutes before* the postgres job failed, so it cannot have been a consequence of it. The matrix also sets `fail-fast: false`. Same thing on the `main` run above: 30m14s. This job is chronically over its cap, independent of this PR.

### ❌ Wrong: "the concurrency group cancels the in-flight run"

There is **no `concurrency:` block in `e2e-tests.yml`** — the only one in the repo is in `helm-publish.yaml`. I asserted a mechanism I hadn't checked.

The real reason `main` doesn't look red: when one job **fails** and another is **cancelled**, GitHub reports the *run-level* conclusion as `cancelled`. So the chronically-timing-out `unit-tests` job masks the genuinely failing postgres job, and every affected `main` run reads as "cancelled" — which scans as an infra hiccup rather than a broken test. **That is why both of these tests merged.** The timeout isn't a cosmetic annoyance; it's suppressing the failure signal on `main`.

### Root cause of the two failures: SQLite is permissive where PostgreSQL is strict

`--db` defaults to `sqlite` (`tests/conftest.py:104`), so that's what runs locally and in the fast job. Both failures are invalid writes that SQLite accepts and PostgreSQL rejects:

**1. Foreign keys are never enforced under SQLite.** `PRAGMA foreign_keys=ON` appears nowhere in the repo, and SQLite defaults it *off*. `_seed_run` (`tests/e2e/rbac/test_rbac_evals.py:42`) writes:

```python
head_completion_id=str(uuid.uuid4()),
```

a fabricated UUID, into a column declared `ForeignKey('completions.id'), nullable=False` (`app/models/eval.py:140`). No such completion exists. SQLite ignores the constraint; PostgreSQL raises `ForeignKeyViolationError`. This test can never have passed on PostgreSQL. (The codebase already knows about this asymmetry — see the comment at `app/models/data_source_connection_tool.py:33-36` about `ON DELETE CASCADE` not being enforced "on SQLite (no foreign_keys pragma)".)

**2. Aware vs naive datetimes.** `BaseSchema.deleted_at` is `Column(DateTime)` — no `timezone=True` — so it maps to `TIMESTAMP WITHOUT TIME ZONE`. The OIDC test writes `deleted_at=_dt.now(_tz.utc)`, timezone-**aware**. asyncpg refuses aware→naive; SQLite stores datetimes as strings and doesn't care.

Worth noting: every *production* soft-delete writer uses naive `datetime.utcnow()` (`git_service.py:555`, `test_case_service.py:219`, `project_service.py:289`, `oauth_server_service.py:200`, …). The naive column is the established convention and the test is the sole outlier — so the narrow fix is the test, not the schema.

### Suggested fixes (for whoever owns these)

1. `test_rbac_evals.py` — create a real completion in `_seed_run`, or make `head_completion_id` nullable if a result legitimately need not have one.
2. `test_oidc_group_sync.py:781` — `deleted_at=_dt.utcnow()` instead of `_dt.now(_tz.utc)`, matching every other writer.
3. **Enable `PRAGMA foreign_keys=ON` for the SQLite test engine.** This is the leverage: it makes the fast default job catch this whole class of bug instead of deferring it to a 70-minute PostgreSQL job nobody reads.
4. Raise `unit-tests`' `timeout-minutes`, or split the job — while it times out, real failures on `main` keep getting laundered into `cancelled`.

None of this is touched by this PR (3 files: 2 frontend components + a doc), so I'm still not fixing it here.

---

yochze:
## Final correction: the masking theory was wrong. These merged past a visibly-red check.

My last comment claimed the timed-out `unit-tests` job masked the postgres failure at the run level, and that **that** is why these tests reached `main`. I checked the originating PRs and it isn't true — the signal was red and visible on each PR's own check list.

### Chain of custody

| Where | `e2e-tests (postgres)` | Failures in that run |
|---|---|---|
| **PR #929** (added `_seed_run`) | failure | `1 failed` — the FK violation it introduced |
| **PR #932** (added the OIDC test) | failure | inherited FK + its own datetime error |
| **`main` @ `8457877`** | failure | both |
| **PR #935** (this one) | failure | both, unchanged |

Each PR introduced its own failure, showed it red as an individual check on its own run, and was merged. PR #929's log is unambiguous:

```
=== 1 failed, 1131 passed, 38 skipped, 104092 warnings in 4988.80s (1:23:08) ===
FAILED tests/e2e/rbac/test_rbac_evals.py::test_case_and_run_listings_narrow_to_one_agent
```

So the real reason is simply that **`e2e-tests (postgres)` is not a blocking check in practice** — it's been red since #929 and merges continue past it. That's a policy question, not a signal-visibility one, and it needs a different fix than the one I proposed.

The `unit-tests` 30-minute timeout is still real and still chronic (#929, #932, `main`, and this PR all show it hitting the cap at 30m±20s), and it does make run-level conclusions read `cancelled` instead of `failure`. But it is **not** what let these in. I conflated a genuine observation with a causal claim I hadn't tested.

### What still stands

The two technical diagnoses are unchanged, and the error text is itself the evidence:

- **FK violation** — SQLite leaves `PRAGMA foreign_keys` off by default and it's set nowhere in this repo, so the fabricated `head_completion_id=str(uuid.uuid4())` (`test_rbac_evals.py:42`) sails through against a `nullable=False` FK to `completions.id`. The repo already documents this asymmetry at `app/models/data_source_connection_tool.py:33-36` ("not enforced on SQLite (no foreign_keys pragma)").
- **Datetime** — the failing SQL literally binds `$9::TIMESTAMP WITHOUT TIME ZONE` and asyncpg reports "can't subtract offset-naive and offset-aware datetimes"; `BaseSchema.deleted_at` is `Column(DateTime)` with no `timezone=True`, and the OIDC test is the only writer passing an aware value.

Caveat on confidence: I could not execute either test — this container has no backend virtualenv and no `sqlalchemy` — so both diagnoses are read from the CI logs and the source, not reproduced locally.

### Revised priority

1. Decide whether `e2e-tests (postgres)` should block merges. While it doesn't, nothing else here matters — the next PG-only breakage lands the same way.
2. Fix the two tests (real completion in `_seed_run`; naive `utcnow()` in the OIDC test).
3. Turn on `PRAGMA foreign_keys=ON` for the SQLite test engine so the fast job catches this class instead of the 80-minute one.
4. Raise or split `unit-tests` — worth doing on its own merits, just not load-bearing for this.

Still not fixing any of it in this PR (2 frontend components + a doc).

---

PR #988, opened 2026-08-17
Safe temperature handling + catalog auto-add for customer providers

yochze:
## Problem

Two related gaps in how deployments track model/provider evolution, both reported from a customer running BOW behind a LiteLLM gateway:

**1. Hardcoded temperature.** The `OpenAi` client — serving the **custom provider** and openai-with-base_url — sent `temperature = 1 if "gpt-5" in model_id else 0.3` on every request. Models behind gateway aliases (Claude Sonnet 5+, gpt-5) reject any non-default temperature with a deterministic 400 (`litellm.UnsupportedParamsError: ... Only temperature=1 is supported`), and aliases defeat any name-based detection.

**2. Frozen model lists.** New models added to `LLM_MODEL_DETAILS` in a release never reached existing deployments: the `GET /llm/models` catalog sync only ran for `is_preset` providers, and customer-created (bring-your-own-key) providers aren't preset — so orgs stayed frozen at whatever models existed when their provider was created.

## Change

**Temperature (commits 1–3)**
- The OpenAI-compatible client sends **no temperature at all** unless one is explicitly configured; the endpoint's default applies.
- New `POST /llm/models/{id}/set_temperature` (manage_llm-gated), stored on `LLMModel.config['temperature']` (survives catalog re-syncs, no migration). Resolution: model config → provider `additional_config` → unset. `0` is a valid explicit value; out-of-range rejected.
- UI: tri-state Temperature field in the model card and Add Model modals — never pre-filled, "Provider default" placeholder when empty, reset arrow to clear. i18n keys in all ten locales.
- Unset keeps each client's historical posture; the Anthropic `_accepts_temperature` gate stays authoritative; reasoning models never receive the param; Bedrock unchanged.

**Catalog auto-add (commit 5)**
- The catalog sync now runs for every provider whose `provider_type` has catalog entries — customer-created providers included — with customer-managed semantics:
  - New catalog models are **auto-added, enabled**, on the next models-list load after an upgrade.
  - Admin choices win: disabled models stay disabled, soft-deleted models are tombstones (never resurrected), and org defaults are never re-pointed — except as a safety net when the org would otherwise have no enabled default (e.g. its default's model id was retired).
  - Retired catalog ids are disabled so orgs don't stay pinned to dead models; no-catalog provider types ("custom") are untouched.

## Verification

- **Live** against Anthropic's OpenAI-compatible endpoint with `claude-sonnet-5`: reproduced the temperature 400, confirmed the fixed client on non-streaming, streaming tool-call, and explicit-temperature paths.
- **Live sandbox** (backend + frontend + real Anthropic API, driven via Playwright): temperature field verified end-to-end (empty start, save, persist, out-of-range rejection, reset, real chat with override active, DB + client-construction checks); catalog auto-add verified through the UI — a provider created with only Claude 4.5 Haiku gained Fable 5 / Sonnet 5 / Opus 5 / Opus 4.8 enabled with catalog pricing on the next page load, Haiku stayed default, and a deleted model stayed deleted across reloads.
- **Tests**: unit suite for client temperature resolution; e2e suites for the temperature override (set/persist/clear/zero/range/permission) and catalog auto-add (auto-add with defaults intact, custom provider untouched, delete/disable choices surviving repeated re-syncs). Mirrors the existing context-window override suite.

Note: `unit-tests` CI carries 13 pre-existing failures that reproduce identically on `origin/main` (see PR comment); this branch's suites pass.

https://claude.ai/code/session_018mFPsDL13yq2mtJ3i2jXP6


yochze:
CI status note: `unit-tests` is red with 13 failures (`test_notification_service`, `test_notify_service`, `test_data_source_member_email`, `test_eval_draft_helpers`, `test_whatsapp_webhook_route`, `test_mcp_tool_registry`) — all reproduce identically on `origin/main`, so they're pre-existing base-branch failures, not from this diff. This PR's own suites (`test_openai_client_temperature`, `test_llm_temperature_override`) pass. The earlier `integration-llms` failure was a runner infra error (couldn't download `dorny/paths-filter`, 429/503) and was retriggered. I'll re-sync when the base branch is green again.

---

PR #998, opened 2026-08-20
Feature/hide data tab

r0533130357-dev:
Let report owners hide the Data tab from shared-report viewers

Sharing a report meant sharing every query behind it: /r/{id} always rendered
the Data tab with the full per-query execution history. An "Include Data Tab"
checkbox in the artifact share dialog now lets the owner send just the report.

- New reports.include_data_tab, defaulting to true (migration shrdata01,
  server_default=true), so every existing report keeps the tab it shows today.
  The frontend treats anything but an explicit false as visible, which also
  covers the window before the report resolves.
- Artifact sharing only — the conversation share has no such tab, so a
  conversation request carrying the field is ignored rather than writing a
  setting that means nothing there.
- Optional on ReportVisibilityUpdate: omitted means leave unchanged, same as
  run_identity. The checkbox gets its own write path in ShareModal instead of
  riding along with saveVisibility, so changing visibility or inviting someone
  can't overwrite the setting with a stale client value.
- /r/{id} watches the flag: a viewer who already opened the Data tab before
  the report loaded is moved back to the report view rather than left on a
  panel whose tab just disappeared.
- Fixes share-event noise this surfaced: set_visibility emitted a "shared
  with ..." session event unconditionally, so every display-only toggle
  (include_data_tab, and run_identity before it) posted one into the report's
  conversation. Now gated on sharing actually changing — visibility differs,
  or a principal list was sent. In-app notifications and emails are untouched.
- share.includeDataTab / includeDataTabDesc added to all ten enabled locales,
  reusing each catalog's existing term for "tab" and its quote convention.

Not a data boundary: the public queries/step endpoints serve the same payload
either way, since they also feed the artifact iframe. This hides the tab, it
does not restrict access.

Tests: 4 new cases in tests/e2e/test_report_sharing.py cover default-on,
round-trip both ways, omitted-leaves-unchanged, conversation-share-ignores-it,
and no-event-on-toggle. 28/28 pass on --db=sqlite; postgres not run locally.


backend/alembic/versions/shrdata01_add_report_include_data_tab.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'shrdata01'
+down_revision: Union[str, None] = 'evalsuitefk01'
```
yochze: **Merge-blocking: this forks alembic history.** Since this branch was cut, `pricovr01` (pricing override on llm_models) landed on `main` with `down_revision = 'evalsuitefk01'` as well. Once this merges, alembic sees two heads descending from `evalsuitefk01` and `alembic upgrade head` aborts with "Multiple head revisions are present" — blocking every deploy and fresh install until re-parented.

Fix: rebase onto current `main` and change this to `down_revision = 'pricovr01'`. This is likely also why the PR's merge state shows unstable.

---

locales/en.json
```diff
@@ -2968,7 +2968,9 @@
     "sharingDisabled": "Sharing disabled",
     "sharingFailed": "Failed to update sharing",
     "userNotFound": "User not found in organization",
-    "copyFailed": "Failed to copy"
+    "copyFailed": "Failed to copy",
+    "includeDataTab": "Include Data Tab",
```
yochze, line 2972: The new `share.includeDataTab` / `share.includeDataTabDesc` keys only landed in 3 of the 10 locale catalogs (en, es, he) — `ar`, `de`, `fr`, `it`, `pt`, `ru`, and `sv` are missing them, though the PR description says all ten were updated. Users in those locales will see English fallback text (or the raw key, depending on fallback config) in an otherwise localized share dialog.

---

yochze (commented):
Reviewed the full diff against the merge base. The feature itself is solid — the omitted-means-unchanged contract, the conversation-share guard, the share-event gating, the separate frontend write path for the checkbox, and the pre-load race handling on /r/{id} all check out, and the new e2e tests cover the right cases. Two issues inline: the alembic migration forks history against current main (merge-blocking), and the new i18n keys are missing from 7 of the 10 locales.

---

PR #1007, opened 2026-08-22
monday: robust board discovery — no more silently missing boards

yochze:
## Why

Customer report: account with 200 boards, only 190 synced — the missing boards active, public, owned by the customer, on a brand-new connection. Investigation (validated live against api.monday.com on a trial account seeded to 211 boards) found four independent ways `MondayClient` board discovery silently loses boards.

## What

All in `backend/app/data_sources/clients/monday_client.py`:

1. **Pagination stops only on an empty page.** Discovery used to break as soon as one page returned fewer than 50 boards; monday does not guarantee non-final pages are full, so a short page mid-crawl truncated every board created after it.
2. **Per-workspace recovery sweep (deep discovery).** The account-wide `boards (limit, page)` listing can omit boards that workspace-scoped or by-id queries return (observed in the wild with shareable boards and cross-product workspaces; monday's own MCP server enumerates per workspace instead of global-crawling). After the global crawl, schema discovery now enumerates `workspaces (membership_kind: all)`, sweeps each workspace's board ids, and recovers anything missing via `boards (ids:)` — with a warning log naming the recovered ids. Query-time board resolution stays shallow (one crawl) and escalates to the sweep only when a board isn't found, so per-query latency doesn't grow with workspace count.
3. **Subitem boards filtered by API `type`, not name.** The `"Subitems of "` name-prefix filter dropped *real* boards monday itself names that way ("expand subitems into a new board") and missed renamed/localized shadow boards. The filter now keys on `type == sub_items_board`; the prefix remains only as a fallback when `type` is absent.
4. **Main-workspace scoping fixed.** Boards in monday's legacy Main workspace return `workspace: null` (documented), so a `workspaces` config filter could never match them. The aliases "Main workspace" / "main" now select them (config field description updated).

Also: the `MAX_BOARDS` cap now counts real boards and logs a warning when it truncates instead of stopping silently, and every discovery run logs a summary (catalog size, per-page sizes, sweep recoveries) so future missing-board reports are diagnosable from backend logs alone.

## Validation

- **Unit**: `tests/unit/test_monday_client.py` — 20 pass, including new regressions for short non-final pages, global-listing omissions recovered via the sweep, type-based subitem filtering (real "Subitems of …" board kept, renamed shadow dropped, legacy no-type fallback), Main-workspace scoping, workspace-listing permission failure, and shallow→deep escalation in `execute_query`.
- **Live** (seeded monday trial, 211 boards, 2 workspaces, private + subitem-named + shadow boards): `get_schemas()` returns all 211; shadow boards excluded; workspace scoping exact; `execute_query` works both shallow and post-escalation.
- **Sandbox e2e** (`sandbox-feedback-loop`): backend + frontend booted, monday data source created with the trial token, `test_connection` OK, `refresh_schema` persisted 211 `datasource_tables` rows, Tables UI showed 0/211 → Select all → **211/211 active** (verified in sqlite). The LLM chat leg was blocked by the sandbox Anthropic key's credit balance (billing 400 — unrelated to this change).
- **Full unit suite**: 15 failures in this sandbox, byte-identical on a clean `origin/main` worktree (notifications, sybase, whatsapp, evals, compaction, artifacts) — pre-existing/environmental, none introduced by this change.

## Out of scope (documented in `docs/feedback-loops/monday-connector.md`)

`save_or_update_tables`/`refresh_schema` key tables by NAME, so a board rename (or a second board taking the same name) deactivates the in-use table and inserts its replacement inactive. The client persists `metadata_json.board_id`; keying monday sync on it is a separate follow-up.

https://claude.ai/code/session_01SxfBPeF6g7yW9MBfwQ4tcA

---
_Generated by [Claude Code](https://claude.ai/code/session_01SxfBPeF6g7yW9MBfwQ4tcA)_

### DIFF
--- a/backend/app/data_sources/clients/monday_client.py
+++ b/backend/app/data_sources/clients/monday_client.py
@@ -16,6 +16,7 @@ label text in `compare_value` to indices using the board's column settings so
 generated queries can filter by the human-readable label.
 """
 import json
+import logging
 import re
 import time
 from typing import Any, Dict, List, Optional
@@ -25,6 +26,8 @@ import requests
 
 from app.data_sources.clients.base import DataSourceClient
 from app.ai.prompt_formatters import ForeignKey, Table, TableColumn, ServiceFormatter
+
+logger = logging.getLogger(__name__)
 
 API_URL = "https://api.monday.com/v2"
 # Multi-level boards first became queryable in 2025-10. Pin the oldest current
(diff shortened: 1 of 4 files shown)


yochze:
Note on the red `unit-tests` check: the 15 failures are pre-existing and unrelated to this diff — they reproduce byte-identically on a clean `origin/main` checkout. They're stale tests referencing code that no longer exists on main (e.g. `NotificationService._build_subject`/`_build_html` were removed, and `completion_feedback_service` imports `LLMService` inside functions so module-level patching fails; same class of staleness in the eval-draft, whatsapp-webhook, notify, mcp-registry, compaction, and artifact-loading tests). The job is `continue-on-error: true`, which is why main shows green while carrying them.

This run's summary is 15 failed / 3657 passed, including all 20 monday connector tests (6 of them new regressions added here). Fixing the 15 stale tests spans six unrelated services, so it belongs in its own cleanup PR rather than widening this one — happy to take that on separately.

---

PR #1010, opened 2026-08-23
Export a dashboard as one standalone, fully offline HTML file

r0533130357-dev:
## Why

A dashboard could leave the product as PPTX (decks), PDF (a flat render) or Markdown (docs) — but there was no way to hand someone **the dashboard itself**: interactive, no login, no network.

The runtime was already most of the way there. `buildArtifactIframeHtml` inlines the data and pulls the libs from `/libs/*.js`, and `get_inline_scripts` already inlines those same libs for headless rendering. This closes the remaining gap.

## What

**Routes** — `GET /artifacts/{id}/export/html` (in-app) and `GET /r/{id}/export_html` (public share page). Both reuse the PPTX route's gates: RBAC, the withheld-snapshot guard, and an `artifact.exported` audit entry. The public one is stricter, because it serves anonymous viewers: it enforces `artifact_visibility` via `get_public_artifacts`, and validates `artifact_id` against what the viewer was just listed rather than trusting it off the request.

**`html_export_service`** — inlines every vendored lib, turns embedded images/PDFs into `data:` URIs, and strips any remote reference left in generated markup (Google Fonts, `<script src>`, `@import`, remote ``), logging what it removed. pdf.js is the one library with an absolute URL baked into our runtime: `artifact-globals.js` assigns `workerSrc = '/libs/pdf.worker.min.js'` during render, so the worker is inlined as a `blob:` URL and pinned with a **getter** — a plain assignment loses that race.

**`jsx_transpile`** — runs the vendored `babel-standalone` inside a headless page (no new dependency) so babel itself can be dropped from the download: **1.64MB instead of ~4MB**. Cached by content hash, with a per-source lock so concurrent viewers don't each launch a Chromium. Falls back to inlining babel if no browser is usable, so the export never hard-fails.

**`artifact-offline-host.js`** — answers the param messages that would otherwise hang. In a standalone file `window.parent === window`, so `post()` does not throw: the message goes to the page itself, nobody replies, and every param control spins forever. `useFilters` is client-side already and needed nothing.

**`artifact_payload`** — extracts the `ARTIFACT_DATA` assembly out of `report_pdf_service` so PDF and HTML share one implementation. The visualization ordering is load-bearing (generated code indexes `viz[0]`, `viz[1]`) and a drifting second copy would bind every chart to the wrong data while still looking plausible. Net −108 lines in the PDF service.

## The one judgment call

Params filter locally **only when they had no value at export time**. The embedded rows are the slice the server already returned, so re-filtering them to a different value would show an empty or partial result as if it were fact. Those controls stay inert and are named in the corner badge — a control that visibly does nothing beats one that quietly lies.

## Validation

`tests/e2e/test_html_export_offline.py` — 16 tests. The core ones open a real export in Chromium over `file://` with every http(s) request **aborted and recorded**, then assert zero remote requests, a drawn chart canvas, Tailwind applied, and filters working offline. Also covered: the inert-param contract, remote-reference stripping, the pdf.js worker pin, RTL/Cyrillic download filenames, and the three public-route gates.

23 passed across the export + PDF suites; the PDF tests confirm the shared-payload refactor is behaviour-preserving. `test_artifact_relationship_loading::test_active_artifact_lookup_does_not_hydrate_report_graph` fails on `main` too — pre-existing, untouched.

## Notes

- `frontend/public/libs/artifact-offline-host.js` is committed to the repo (a `.gitignore` exception alongside `artifact-globals.js`); the Docker build already copies `frontend/public/libs`, so no packaging change.
- Page-mode artifacts only. Docs render in Vue and decks are python-pptx source — neither has a React runtime to stand up.
- Follow-up worth considering: the codegen prompt still permits remote assets. The export strips them, but a prompt rule would stop them at the source.


backend/app/services/artifact_payload.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # Applied values start at each param's default, exactly as the live host
+    # seeds them. Identity params carry no client value — the server binds
+    # them per viewer, and an export has no viewer.
+    values: dict[str, Any] = {}
+    for name, entry in by_name.items():
+        if entry["source"] == "identity":
+            continue
+        if entry["default"] is not None:
+            values[name] = entry["default"]
```
yochze, line 249: **`values` is seeded from `spec.default` only, ignoring the exported step's `applied_params`.**

The comment says "Applied values start at each param's default, exactly as the live host seeds them" — but the live host is then driven by the user, whereas the export bakes in a specific *snapshot* that was produced under whatever params were actually applied when the step ran.

Consider a param `country` with no default, where the owner ran the query with `country='US'` (stored on the step as `applied_params`). Here `values` ends up empty for `country`, so `artifact-offline-host.js` classifies it as *emulatable* (its baked-in check reads `baseParams.values`). An offline viewer then picks `country='DE'`, the host re-filters the US-only snapshot rows client-side, and silently presents an empty/partial result as fact — exactly the case the offline host's own header comment says must stay inert.

Seed `values` from the exported step(s)' `applied_params`, falling back to `spec.default`, so a param baked into the snapshot is correctly marked baked-in rather than emulatable.

---

backend/app/services/html_export_service.py
```diff
@@ -0,0 +1,273 @@
...
+_REMOTE_PATTERNS: tuple[tuple[str, re.Pattern], ...] = (
+    ("stylesheet link", re.compile(r"""<link\b[^>]*\bhref=["']https?://[^"']*["'][^>]*>""", re.I)),
+    ("script src", re.compile(r"""<script\b[^>]*\bsrc=["']https?://[^"']*["'][^>]*>\s*</script>""", re.I)),
+    ("css @import", re.compile(r"""@import\s+(?:url\()?["']?https?://[^"')]+["']?\)?\s*;""", re.I)),
+)
+
+# Kept separate: rewriting an <img> to nothing would collapse a layout, so the
+# src is blanked in place and the alt text carries the meaning.
+_REMOTE_IMG = re.compile(r"""(<img\b[^>]*\bsrc=)["']https?://[^"']*["']""", re.I)
+
```
yochze, line 59: **The remote-reference stripping leaves several URL forms through, so the "fully offline" file can still make network requests.**

`_REMOTE_PATTERNS` / `_REMOTE_IMG` only match `<link href>`, `<script src>`, `@import`, and ``. They miss:

- **Inline-style `url()`** — `<div style="background-image:url('https://tracker.example/p.gif')">`
- **`<style>` block `url()`** that isn't an `@import` (e.g. `background:url(https://…)`)
- **`srcset`** — ``'&lt;img srcset="https://cdn.example/x.png 2x"&gt;'`` (only `src=` is blanked)
- **media `src`** on `<video>`/`<audio>`/`<source>`/`<iframe>`
- **protocol-relative URLs** — `//host/...` (the `https?://` anchor doesn't match)

Any of these fires a live request on open — effectively a tracking beacon revealing when/where the snapshot is viewed — in a file the PR promises makes zero network requests. Worth broadening the patterns (and/or a CSP `<meta>` `default-src 'self' data:` as a belt-and-suspenders backstop).

---

backend/app/routes/report.py
```diff
@@ -682,6 +682,85 @@ async def _generate_locked() -> str | None:
     )
 
 
+@router.get("/r/{report_id}/export_html")
+async def export_public_report_html(
```
yochze, line 797: **This public route is missing two gates the in-app route enforces**, so it's a weaker path to the same data:

1. **No `artifact.exported` audit entry.** The in-app `export_artifact_html` logs `action="artifact.exported"` (`artifact.py`), and the PR description states both routes write the audit entry — but this one doesn't. A signed-in member who downloads the full snapshot via the share link instead of the in-app button leaves no audit trail, so a "who exported this dashboard" review silently misses every share-page export.

2. **No `artifact.status == "failed"` guard.** The in-app route returns 400 ("Regenerate it before exporting") when the artifact failed to generate; here an anonymous viewer gets a successfully-downloaded standalone file built from stale/broken `content.code` with no warning.

Both checks exist a few lines up in the in-app twin — worth mirroring them here.

---

backend/app/routes/report.py
```diff
@@ -682,6 +682,85 @@ async def _generate_locked() -> str | None:
...
+            status_code=409,
+            detail="This dashboard runs with each viewer's own credentials, so it cannot be exported.",
+        )
+
+    artifact = await db.get(ArtifactModel, chosen_id)
+    if artifact is None or artifact.deleted_at is not None:
+        raise HTTPException(status_code=404, detail="Not found")
+
+    try:
+        html = await build_standalone_html(db, artifact)
```
yochze, line 885: **Unbounded headless-Chromium work on an anonymous route.**

`build_standalone_html` can launch a headless Chromium for JSX transpile on a cache miss. Unlike the sibling public PDF route (`export_public_report_pdf`), which wraps the same class of work in `_pdf_export_locks` (a per-report lock) plus a `wait_for` timeout / 504, this route has no concurrency cap and no overall timeout — and it's reachable unauthenticated.

An anonymous client requesting exports of N different public dashboards (or varying uncached artifact code) concurrently gets N parallel browser launches — the per-source transpile lock only dedupes *identical* source — which can exhaust memory and stall the API on a small instance. Consider reusing the PDF route's lock + timeout pattern here.

---

frontend/components/dashboard/ArtifactFrame.vue
```diff
@@ -710,6 +723,94 @@ async function exportPptx() {
...
+    // matters more here — the payload IS html, and handing the browser a
+    // remote text/html URL to open would run it in our origin.
+    const arrayBuffer = await response.arrayBuffer();
+    const localBlob = new Blob([arrayBuffer], { type: 'text/html' });
+
+    const rawTitle = selectedArtifact.value?.title || 'dashboard';
+    const fallbackName =
+      `${String(rawTitle).replace(/[^\w\s.-]/g, '').trim().slice(0, 120) || 'dashboard'}.html`;
+    const filename =
+      filenameFromDisposition(response.headers.get('Content-Disposition')) || fallbackName;
```
yochze, line 844: **`Content-Disposition` isn't readable cross-origin, so this filename-parsing path is dead code in split-origin deployments.**

`Content-Disposition` is not a CORS-safelisted response header, and `backend/app/core/cors.py` sets no `expose_headers` on the `CORSMiddleware`. When `BOW_CORS_ALLOWED_ORIGINS` is set (frontend served from a different origin than the API), `response.headers.get('Content-Disposition')` returns `null`, so `filenameFromDisposition(...)` always yields `null` and the download silently falls back to the locally-stripped `fallbackName`.

The upshot: the server's careful RFC 5987 / transliterated-filename work never applies precisely where CORS is in use — a Hebrew-titled dashboard downloads under the stripped ASCII fallback instead of its real name. Fix is one line on the backend: add `expose_headers=["Content-Disposition"]` to the `CORSMiddleware` config.

---

yochze (commented):
Reviewed the standalone offline HTML export. The design is solid — reusing the PPTX route's gates, extracting the shared payload builder, and the offline-host data-ingest wiring all check out, and the RFC 5987 filename handling and `</script>`/U+2028 escaping are nicely done.

Five findings inline, grouped by theme:

**Public share route is a weaker copy of the in-app route** — it omits the `artifact.exported` audit entry (the PR description says both write it) and the `status == "failed"` guard, and unlike the sibling public PDF route it has no concurrency lock / timeout around work that can launch headless Chromium for anonymous callers.

**"Fully offline" has gaps** — the remote-reference stripping misses inline-style `url()`, `srcset`, media `src`, and protocol-relative URLs, so an exported file can still fire live requests.

**Two edge-case correctness issues** — exported `params.values` is seeded only from declaration defaults (not the step's `applied_params`), so the offline host lets viewers "emulate" a param that was actually baked into the snapshot; and the new `Content-Disposition` filename parsing is dead code in split-origin deployments because the backend never adds `expose_headers`.

None are blockers for the core feature, but the public-route parity gaps and the offline-stripping gaps are worth addressing before merge given this file leaves our control entirely once downloaded.

---

PR #1011, opened 2026-08-23
Feature/reports UI

r0533130357-dev:
## Align /projects with the /reports design system

Rebuilds the projects index on the /reports layout and fills two gaps in the
reports list.

**/projects**
- Scope tabs: My projects / Shared with me (split on `is_owner`).
- Search over name + description, plus a filters popover. Only the two filters
  a project row actually carries survive from /reports — Access (private / org)
  and Reports (with reports / empty); type, schedule, agent and dashboard have
  no data in the `/projects` payload.
- Pagination: rows-per-page (10/25/50/100, persisted) + page arrows.
- Rows carry no actions — no bulk select, no archive. Everything else matches
  /reports: bordered card, folder avatar in the project accent, metadata
  sub-line, date column, loading skeleton, empty states.
- All filtering/paging is client-side: `GET /projects` already returns every
  visible project in one response, so no backend change.

**/reports**
- Rows now show which project a report lives in — a folder chip in the project
  accent, in the metadata sub-line, linking to the project. Rendered only for
  projects the caller can actually open, so a report shared out of a private
  project doesn't surface a folder name or link into a 403.
- Training sessions are no longer labelled "Chat": `mode === 'training'` wins
  over the artifact type and renders as "Training" with the mode picker's
  academic-cap icon. Previously the Type → Training filter had no visible
  counterpart on the rows.

**Both** — active tab (text + underline) is now `blue-500` (#3b82f6).

New i18n keys added across all 10 locales.


yochze:
**Bug: `reports.sharedEmpty` / `reports.sharedEmptyHint` missing from all 9 non-English locales**

`frontend/pages/reports/index.vue` renders these keys on the "Shared with me" empty state:

```
L427: {{ activeFilter === 'shared' ? $t('reports.sharedEmpty') : $t('reports.empty') }}
L435: {{ activeFilter === 'shared' ? $t('reports.sharedEmptyHint') : $t('reports.emptyHint') }}
```

But those two keys were only added to `en.json`. The parallel `projects.*` pair was added to every locale, while the `reports.*` pair was forgotten everywhere except English:

- `en.json` — `sharedEmpty`/`sharedEmptyHint` present in **both** the `reports` block (L703) and `projects` block (L851)
- `de.json`, `ar.json`, `es.json`, `fr.json`, `he.json`, `it.json`, `pt.json`, `ru.json`, `sv.json` — present **only** in the `projects` block

**Impact:** a non-English user opening `/reports` → "Shared with me" with nothing shared sees the English fallback strings on an otherwise-translated page.

**Fix:** add `reports.sharedEmpty` and `reports.sharedEmptyHint` to the 9 non-English locale files (the `reports` block).

---

PR #1013, opened 2026-08-23
monday: id-based joins for connected boards

yochze:
## Why

Customer request: first-class Connected Boards support ("a lookup column to an item on another board"). After #1012, connect-boards columns return the linked items' *display names* — but joining boards on names is ambiguous (names repeat) and multi-link cells are comma-joined strings. The monday API exposes `linked_item_ids` on `BoardRelationValue` and `DependencyValue`; this PR surfaces it so cross-board joins run on ids.

## What

`backend/app/data_sources/clients/monday_client.py`:
- Every connect-boards / dependency column now emits a companion **`"<Column> (item_ids)"`** DataFrame column holding the linked item ids as a `list[int]` (fragments now request `display_value linked_item_ids`).
- The published schema lists the companion column, and the `board_relation` **foreign key now points from the companion to the linked table's `item_id`** — the id column is the joinable one the FK semantics always meant.
- Selecting `"<Column> (item_ids)"` in a query spec resolves to its parent column (the companion is emitted alongside it).
- The query system prompt documents the join recipe — explode the companion, merge on the linked board's `item_id` — with a worked example, and notes mirror columns remain display-text only (the API exposes no ids for mirrors).

## Validation

- **Unit**: `tests/unit/test_monday_client.py` — 22 pass, including new regressions for the companion column + retargeted FK in the schema, ids in query results, companion-name selection, and the documented explode-and-merge join producing correct row multiplicity.
- **Live** (monday trial account, two linked boards): schema carries `Linked Account (item_ids)` with the FK targeting the linked table; query results carry the real linked id; the exact prompt-documented join returns the linked item (`Task 1 → "Linked target item"`).

Follow-up to #1007 / #1012.

https://claude.ai/code/session_01SxfBPeF6g7yW9MBfwQ4tcA

---
_Generated by [Claude Code](https://claude.ai/code/session_01SxfBPeF6g7yW9MBfwQ4tcA)_

### DIFF
--- a/backend/app/data_sources/clients/monday_client.py
+++ b/backend/app/data_sources/clients/monday_client.py
@@ -90,6 +90,13 @@ _TYPE_MAP = {
 # Column types whose settings carry an index→label map that rules filter by.
 _LABELED_TYPES = {"status", "dropdown"}
 
+# Column types that link to items and expose `linked_item_ids` — each gets a
+# companion "<Column> (item_ids)" DataFrame column for precise id-based joins
+# (display names are ambiguous: duplicates exist and multi-link cells are
+# comma-joined).
+_LINKED_ID_TYPES = {"board_relation", "dependency"}
+_ITEM_IDS_SUFFIX = " (item_ids)"
+
 
 class MondayClient(DataSourceClient):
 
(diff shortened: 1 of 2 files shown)


yochze:
Note on the `e2e-tests (sqlite)` failure reported on this branch's previous head: it was **not** from the monday change. The offline-HTML export route (`export_public_report_html`, from the export-dashboard-as-HTML feature merged in #1010) gained a required `request: Request` parameter, but the public-export e2e helper in `tests/e2e/test_html_export_offline.py` still called the route without it — failing `test_public_export_is_refused_when_the_dashboard_is_not_shared` and `test_public_export_rejects_an_artifact_from_another_report` on every run since, including against main. (#1010 and #1012 were merged before their ~50-minute e2e runs completed, so it surfaced only now.)

Since it keeps the e2e suite red for every PR, this branch carries a scoped one-commit fix (2e41619): the helper now passes a minimal ASGI-scope `Request` (the route uses it only for best-effort audit logging). Verified locally: the two tests pass, and the whole file is 9 passed / 7 skipped.

---

PR #1018, opened 2026-08-24
edit the name of the llm and ui

r0533130357-dev:
Custom models (Azure deployments, Bedrock, OpenAI-compatible endpoints, fine-tuned ids) were always named after their model_id — the frontend duplicated it into name on every creation path. This makes the display name editable.

What changed

PATCH /llm/models/{id} now renames custom models. Preset models are rejected with a 400: their name is owned by LLM_MODEL_DETAILS and is re-applied by _sync_provider_with_latest_models on every models-list load, so accepting a rename would only look like it worked. Blank resets to the model_id.
Both creation paths (POST /llm/models and the models array in POST /llm/providers) honor a supplied name and fall back to the model_id. This also fixes an IntegrityError for API callers that sent only a model_id — name is NOT NULL and _create_models had no fallback.
UI: a Display Name field in the model card (custom models only) and in both add-model dialogs. New locale keys across all 10 languages.


yochze (commented):
## Review

Solid, well-engineered change — the editable-display-name feature is correct, defensive, and thoroughly tested. Main feedback is about packaging, not correctness.

### Strengths

**Backend is careful and correct.**
- Rejecting preset renames with a 400 is the right call — `_apply_catalog_model_details` re-applies catalog names on every `get_models` load, so accepting would silently revert. Blank resets to `model_id`, respecting the NOT NULL column.
- Adding `response_model=LLMModelSchema` to the PATCH fixes a real latent bug: serializing the raw ORM row would hit `RecursionError` on the `provider → models → provider` cycle. Using the non-`WithProvider` schema avoids it.
- The `name` fallback in `_create_models` fixes a genuine `IntegrityError` for API callers that sent only a `model_id`. Verified that an explicit name still wins over the catalog "nicer label" upgrade (the catalog branch only overrides when `not model.get("name")`).
- `model_router._match` and `test_suite_service._resolve_model_slug` both resolve **model_id first, display name last** — good foresight against the new failure mode where an editable, non-unique name could shadow another model's id.

**Frontend is consistent** across all three entry points (add-model modal, model card, provider dialog), gated to `is_custom`, with correct dirty-state/reset semantics.

**Tests are excellent** — `test_llm_model_rename.py` covers rename persistence across re-sync, blank reset, preset rejection, both creation paths, and the permission gate.

**i18n parity verified** — all 10 locales have 0 keys missing vs `en.json`, all JSON parses.

### Non-blocking observations

1. **Scope & title.** This bundles a large reports-UI i18n/RTL externalization sweep (~2,080 of the +2,645 lines, across the 10 locale files) with the LLM-naming feature, under the title *"edit the name of the llm and ui"* (commit message typo: *"mml"*). Two separate PRs/commits — or at least a clearer title — would make the history much more useful.
2. **Name-based routing is a last-resort alias.** `_match` resolving by display name is safe today (the tool schema only offers `model_id`), but display names are editable and non-unique, so two custom models sharing one would resolve ambiguously. Already acknowledged in the comment; low risk.
3. **No backend length cap** on `name` (frontend caps at 120). Typical for this codebase, not a real concern.

### CI

`unit-tests` is red, but **all 14 failures are in unrelated modules this PR doesn't touch** (`test_notification_service`, `test_whatsapp_webhook_route`, `test_notify_service`, `test_mcp_tool_registry`, `test_eval_draft_helpers`, `test_data_source_member_email`, `test_artifact_relationship_loading`) — pre-existing failures, not caused by this change. The PR's own **e2e (postgres + sqlite), playwright, and both integration jobs all passed.** Worth noting the branch still can't merge on a required-green gate until those unrelated failures are fixed on `main`.

---

PR #1020, opened 2026-08-25
Support Azure AI Foundry endpoints & edit model id

r0533130357-dev:
Closes the report that only Azure OpenAI endpoints could be connected. Foundry resources
now work natively in the Azure connector — no `custom` provider workaround.

**Why it was broken:** the Azure provider always built the Azure SDK client, which calls
`/openai/deployments/{name}/chat/completions?api-version=`. Foundry resources
(`*.services.ai.azure.com`) don't serve that route at all.

### What changed

- **Endpoint routing** — the surface is resolved from the endpoint (hostname, or an explicit
  `endpoint_mode` for private DNS / Private Link), then the client follows:
  - Azure OpenAI → Azure SDK, unchanged
  - Foundry + OpenAI deployment → `/openai/v1`
  - Foundry + Anthropic deployment → `/anthropic/v1` (Messages API)
- **Portal URLs are normalized.** The Foundry portal shows the *project* endpoint
  (`.../api/projects/<name>`); appending `/openai/v1` to it produced a path APIM answers
  with a 401 blaming the API key — undebuggable from the UI.
- **`Anthropic(base_url=...)` is honored.** It was accepted and silently dropped, pinning
  every caller to `api.anthropic.com`.
- **404s are no longer retried.** `DeploymentNotFound` / `api_not_supported` classify as a
  new permanent `model_not_found` (fallback-eligible at model scope) instead of the
  retryable `provider_error`. A missing deployment used to burn the full retry budget across
  the LLM facade, planner and fallback chain and surface as ~45s of silent "thinking";
  it now fails in ~1s with a clear message. Applies to every provider.
- **UI** — segmented Endpoint type control with a live "Detected: …" badge, so auto-detection
  shows its work before you save.

### Notes

- No migrations. Everything lives in the provider's `additional_config` JSON; existing
  Azure providers and `custom`-provider workarounds keep working untouched.
- Anthropic deployments are matched by name (`claude` / `anthropic`). A deployment renamed
  to hide the family lands on the OpenAI surface with a clear `model_not_found`, not a stall.
- Not in scope: Entra ID / Managed Identity auth (tracked separately) — this is API-key only.


backend/app/ai/llm/llm.py
```diff
@@ -223,19 +306,30 @@ def __init__(
...
             return base + "/v1/"
+        # Strip the non-inference paths the Azure portal hands out, so pasting
+        # one does not silently produce a URL that exists nowhere. The Foundry
+        # portal shows the project endpoint
+        # (``.../api/projects/<name>``) far more prominently than the resource
+        # root, and ``/models`` is the older Azure AI Inference route; appending
+        # /openai/v1 to either yields a path APIM answers with a 401 whose
+        # message blames the key, which is near-impossible to debug from the UI.
+        base = re.sub(r"/api/projects/[^/]+$", "", base)
+        base = re.sub(r"/models$", "", base)
```
yochze: **`_azure_v1_base_url` normalizes the portal's project and `/models` URLs, but not the Target URI the Foundry portal shows for Claude deployments.**

For an Anthropic deployment the portal displays `https://<res>.services.ai.azure.com/anthropic/v1/messages` as the Target URI. Pasting that into `endpoint_url` leaves the path intact, so an Anthropic model calls `.../anthropic/v1/messages/anthropic/v1/messages` and an OpenAI model calls `.../anthropic/v1/messages/openai/v1/chat/completions` — both 404, which is the exact silent-dead-URL failure mode these two `re.sub` lines were added to prevent for the other portal-displayed URLs. (Same applies to a pasted `.../openai/v1` target.)

Suggest also stripping the surface-specific tails, e.g.:

```python
base = re.sub(r"/(anthropic|openai)(/v1(/messages)?)?$", "", base)
```

---

yochze (commented):
One inline finding on the endpoint normalization (pasting the portal's Claude/OpenAI Target URI still produces dead request paths).

---

yochze:
@r0533130357-dev — summary of the review findings on this PR, ordered by importance:

**1. Fixing a model_id doesn't clear the sticky circuit breaker (worth fixing before merge).**
This PR makes `model_not_found` open a sticky breaker (15-min cooldown) keyed by the model's DB row id (`fallback.py`), and ships the model_id edit as the remedy for a typo'd deployment name — but `update_model` never resets that key. Sequence: run hits the bad deployment → breaker opens → admin fixes the id via PATCH → the row id is unchanged, so the model stays routed-around for up to 15 minutes with no indication why, making the fix appear not to work. A best-effort breaker reset in `update_model` when `model_id` changes is a few lines.

**2. Pasting the portal's Claude Target URI still produces dead request paths** — see the inline comment on `_azure_v1_base_url`: it strips `/api/projects/<name>` and `/models` but not `/anthropic/v1/messages` (or `/openai/v1`), so the URL the Foundry portal shows for a Claude deployment yields `.../anthropic/v1/messages/anthropic/v1/messages` → 404. One more `re.sub` closes it.

**3. The claude/anthropic name heuristic is a documented tradeoff — fine to ship as-is.**
The dual-client split itself is unavoidable (verified: Foundry serves Claude only on `/anthropic/v1/messages`; both OpenAI-shaped routes 404). The residual edge is a *classic* Azure OpenAI deployment an admin happened to name "claude-…", which would be rerouted and break with the sticky failure from item 1. Unlikely in practice; an explicit per-model family override (fits naturally next to the new model-edit UI) would eliminate it if you ever want to.

**4. Optional follow-up: the Anthropic SDK now ships a first-party Foundry client.**
The repo pins `anthropic>=0.40.0,<1.0`, which predates `AnthropicFoundry` — that client handles the Foundry URL shape and auth natively and would replace the hand-rolled `base_url` derivation (and most of item 2's normalization concerns). We can upgrade the SDK if needed; it's a bigger change than this PR should absorb, but a good direction after it lands.

---

PR #1021, opened 2026-08-25
rejeact-all-update

r0533130357-dev:
Treatment of the problem of dizziness: instructions and immediate refreshment.

### DIFF
--- a/backend/app/services/instruction_service.py
+++ b/backend/app/services/instruction_service.py
@@ -1537,14 +1537,7 @@ class InstructionService:
         meta_src = None
         if instruction.current_version_id:
             meta_src = await self.version_service.get_version(db, instruction.current_version_id)
-        has_main_build = (await db.execute(
-            select(InstructionBuild.id).where(
-                InstructionBuild.organization_id == str(instruction.organization_id),
-                InstructionBuild.is_main == True,  # noqa: E712
-                InstructionBuild.deleted_at == None,  # noqa: E711
-            ).limit(1)
-        )).scalar_one_or_none()
-        if has_main_build:
+        if await self._org_has_main_build(db, instruction.organization_id):
             # The org HAS a main build and this instruction isn't in it: a NEW
             # instruction that so far exists only in pending suggestion builds.
             # Its live baseline is EMPTY text. The row/current-version cache
@@ -2204,6 +2197,23 @@ class InstructionService:
                 return r.get("proposed_version_id") == pv
         return False
 
+    async def _org_has_main_build(self, db: AsyncSession, organization_id) -> bool:
+        """True when the org keeps its live content in a main build at all.
+
+        False for orgs that predate main-build content, where the cached
+        instruction row / current version IS the live text. In such an org
+        `_is_in_main_build` is False for EVERY instruction, live ones included,
+        so its answer proves nothing on its own.
+        """
+        from app.models.instruction_build import InstructionBuild
+        return (await db.execute(
+            select(InstructionBuild.id).where(
+                InstructionBuild.organization_id == str(organization_id),
+                InstructionBuild.is_main == True,  # noqa: E712
+                InstructionBuild.deleted_at == None,  # noqa: E711
+            ).limit(1)
+        )).scalar_one_or_none() is not None
+
     async def _is_in_main_build(self, db: AsyncSession, instruction) -> bool:
         """True when the live build carries this instruction."""
         from app.models.instruction_build import InstructionBuild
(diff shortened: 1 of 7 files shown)


backend/app/services/instruction_service.py
```diff
@@ -2204,6 +2204,60 @@ def _settled_marker_matches(build, instruction_id: str,
...
+        if not rows:
+            # No proposal to have rejected — leave unexplained state alone.
+            return False
+        for build, _proposed_text, proposed_vid in rows:
+            if self._voided_marker_matches(build, iid):
+                continue
+            pv = str(proposed_vid) if proposed_vid else None
+            if not self._settled_marker_matches(build, iid, pv):
+                return False  # still on the table somewhere
+        instruction.deleted_at = datetime.utcnow()
```
yochze, line 2304: **Bug: this can soft-delete a LIVE published instruction in a legacy org with no main build.**

The retire decision checks only main-build membership, never `instruction.status`. This file documents the legacy state elsewhere (~line 1558: orgs that predate main-build content, where the cached row / current version IS the live text and `_is_in_main_build()` is `False`). In such an org, if an agent stages an **edit** suggestion against a published instruction and the reviewer rejects all hunks: `_settle_resolved_suggestion_rows` stamps the only proposal settled, `_is_in_main_build` returns `False` (no main build exists), and `instruction.deleted_at` is set — the live instruction is destroyed by rejecting an edit to it.

The backfill script guards this with `Instruction.status == "draft"`; this runtime path should too:

```python
if str(getattr(instruction, "status", "")) != "draft":
    return False
```

Note `test_reject_all_on_an_edit_keeps_the_live_instruction` only covers orgs whose main build carries the instruction, so it doesn't catch this.

---

frontend/components/KnowledgeExplorer.vue
```diff
@@ -2865,9 +2865,22 @@ const mergedSegments = computed(() => {
...
-  const { data } = await useMyFetch<Instruction>(`/api/instructions/${detail.value.id}`, { method: 'GET' })
-  if (data.value) { detail.value = data.value; if (!editing.value) syncDraft(data.value) }
-  await loadPending(detail.value.id); await loadVersions(detail.value.id); refreshLists(); fetchReviewCount()
+  const id = detail.value.id
+  const { data } = await useMyFetch<Instruction>(`/api/instructions/${id}`, { method: 'GET' })
+  // Rejecting the suggestion that would have INTRODUCED an instruction retires
+  // it server-side — there is no row left to reload. Keeping the pane open on
+  // the stale copy would leave the reviewer reading an instruction the tree no
+  // longer lists.
+  if (!data.value) {
```
yochze, line 2876: **Any failed GET is treated as "instruction was retired", not just a 404.**

If the follow-up `GET /api/instructions/{id}` fails transiently (500, network blip, expired auth), `data.value` is null and this branch drops the row from `allInstructions`, closes the detail pane, and clears selection/version/diff state — the reviewer loses their place mid-review even though the instruction still exists.

`useMyFetch` also returns an `error` ref; the destructive cleanup should be gated on the actual 404:

```ts
const { data, error } = await useMyFetch<Instruction>(`/api/instructions/${id}`, { method: 'GET' })
if (!data.value) {
  if (error.value?.statusCode !== 404) return  // transient failure — keep the pane
  ...
}
```

---

tools/agent/retire_rejected_instruction_drafts.py
```diff
@@ -0,0 +1,99 @@
...
+                if not service._voided_marker_matches(build, str(instruction.id))
+                and not service._settled_marker_matches(
+                    build, str(instruction.id),
+                    str(proposed_vid) if proposed_vid else None,
+                )
+            ]
+            if undecided:
+                kept.append((instruction.id, label, f"{len(undecided)} proposal(s) still open"))
+                continue
+            retired.append((instruction.id, label))
```
yochze, line 76: **The script re-implements the retire decision it claims to delegate, and in `--apply` mode discards the helper's return value.**

The docstring says the decision "is delegated to `_retire_rejected_create_suggestion`... so this script cannot drift from the runtime rule", but lines 62–79 duplicate the is-in-main / pending-builds / marker checks to build the report, and then the `--apply` call below ignores the helper's `bool`. If state changes between the script's check and the helper's re-check — or the two implementations diverge (today the helper lacks this script's `status == "draft"` guard) — a row is printed under "Retired" that the helper actually declined.

Simpler and drift-proof: give the helper a `dry_run` flag (or just call it and report its return value), and drop the duplicated decision logic here.

---

yochze (commented):
Reviewed the reject-all/retirement flow. The core design is sound — marker-based verdicts, fail-open search filtering, settle-only-when-no-live-hunks all check out — but three findings, most severe first:

1. **`_retire_rejected_create_suggestion` can soft-delete a live published instruction** in a legacy org with no main build (it checks main-build membership but never `instruction.status`).
2. **`reloadAfterResolve` treats any failed re-fetch as a 404**, so a transient error drops the row and closes the reviewer's pane.
3. The backfill script duplicates the retire decision it claims to delegate and ignores the helper's return value in `--apply` mode.

Details inline. Also verified and cleared: the static marker helpers called at class level, session expiry across the mid-helper commit, partial-reject safety, and the deliberate fail-open in `_decided_draft_ids`.

---

PR #1027, opened 2026-08-27
fix(reports): stop exposing agent/connection internals in report responses

yochze:
## Problem

Report responses (`POST /api/reports`, `GET /api/reports/{id}`) serialized attached data sources straight off the ORM, exposing agent-management metadata to any report creator/viewer — not just agent admins:

- **Connection `config`** via the embedded `ConnectionEmbedded` schema — for MCP connections this includes the server URL, static headers (a common place for API keys when not using the bearer-credential flow), and the header/metadata **injection rules** describing how user identity is forwarded. For an unauthenticated MCP server that trusts identity headers, this is an impersonation recipe.
- **The agent's admin-authored `context`** (prompt-injection text).

`GET /reports/{id}` is additionally reachable by shared-report viewers (`allow_public=True`). Encrypted credentials were never exposed — they live in a separate encrypted column with no response-schema field.

Two adjacent API bugs fixed in the same pass:
- `POST /api/reports` without a `title` returned a raw 500 (`AttributeError` in slug generation) instead of succeeding.
- `ReportCreate.widget` was a dead field: extracted and discarded, never used.

## Changes

- **`ConnectionReportEmbedded`** (new): slim connection view for report responses carrying only the display/auth-state fields the report UI actually reads — `type`, `connector_key`, `auth_policy`, `allowed_user_auth_modes`, `user_status`, `data_shape`, `is_active`. `config` is accepted as an input-only field (`exclude=True`) solely so `connector_key` can still be derived for provider icons; it is never serialized.
- **`DataSourceReportSchema`**: drops `context` and the legacy top-level `config`; `connections` now uses the slim schema. Full config remains available on the admin data-source endpoints via `ConnectionEmbedded`, unchanged.
- **Slug generation** defaults to `untitled` when `title` is missing/null, so report creation succeeds instead of 500ing.
- **`ReportCreate.widget` removed**; unknown `widget` keys from old clients are silently ignored (Pydantic extra-key behavior), verified by test.

No frontend changes needed: report views only read `connections[0].type` / `connector_key` (icons) and `auth_policy` / `user_status` (per-user sign-in flow), all of which are preserved. No backend caller passed `widget=`; AI context builders read ORM objects, not response schemas.

## Tests

- New `tests/e2e/test_report_response_sanitization.py`:
  - create/get/list report responses carry no `config`/`context`/`credentials` on data sources or embedded connections, while display fields (name, type) survive
  - `POST /reports` without `title` (and with `title: null`) returns 200 with an `untitled-*` slug
  - legacy `widget` key in the payload is ignored
- Existing report/data-source e2e + unit suites pass (69 tests): `test_report.py`, `test_report_read_regressions.py`, `test_report_sharing.py`, `test_data_source.py`, `test_report_activity.py`, `test_report_data_source_lifecycle.py`

https://claude.ai/code/session_01WFUY2ToCPUN2M61drwEQLq

---
_Generated by [Claude Code](https://claude.ai/code/session_01WFUY2ToCPUN2M61drwEQLq)_

### DIFF
--- a/backend/app/schemas/data_source_schema.py
+++ b/backend/app/schemas/data_source_schema.py
@@ -171,14 +171,71 @@ class ConnectionEmbedded(BaseModel):
         from_attributes = True
 
 
+class ConnectionReportEmbedded(BaseModel):
+    """Slim connection info for Report responses — display/auth-state fields only.
+
+    Deliberately excludes ``config``: report responses reach every report
+    creator/viewer (including shared-report viewers via /r/{id}), while a
+    connection's config can carry internal topology — server URLs, MCP static
+    headers, identity header-injection rules. Full config stays on the admin
+    data-source endpoints via ConnectionEmbedded.
+
+    ``config`` is still *accepted* here (excluded from serialization) because
+    connector_key must be derived from it when validating straight off ORM
+    objects, same as ConnectionEmbedded.derive_connector_key.
+    """
+    id: str
+    name: str
+    type: str
+    auth_policy: str = "system_only"
+    allowed_user_auth_modes: Optional[List[str]] = None
+    is_active: bool = True
+    user_status: Optional[DataSourceUserStatus] = None
+    connector_key: Optional[str] = None
+    data_shape: str = "tables"
+
+    # Input-only: consumed by the validators below, never serialized.
+    config: Optional[Any] = Field(default=None, exclude=True)
+
+    @validator('allowed_user_auth_modes', pre=True)
+    def parse_json_fields(cls, v):
+        if isinstance(v, str):
+            try:
+                return json.loads(v)
+            except (json.JSONDecodeError, TypeError):
+                return None
+        return v
+
+    @validator('connector_key', always=True)
+    def derive_connector_key(cls, v, values):
+        if v:
+            return v
+        return _connector_key_from_config(values.get('config'))
+
+    @validator('data_shape', always=True)
+    def derive_data_shape(cls, v, values):
+        try:
+            from app.schemas.data_source_registry import data_shape_for
+            return data_shape_for(values.get('type')) if values.get('type') else v
+        except Exception:
+            return v
+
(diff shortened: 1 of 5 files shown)


yochze:
**`unit-tests` failure on this PR is pre-existing on `main`, not from this diff.**

The same 14 tests fail identically on the base branch:
- This PR's run: [job 98454406635](https://github.com/bagofwords1/bagofwords/actions/runs/33053481914/job/98454406635) (head `21d18e6`) — 14 failed.
- `main`'s own latest run on `4c853d9` (this branch's exact base): [job 98431369241](https://github.com/bagofwords1/bagofwords/actions/runs/33046423283/job/98431369241) — `unit-tests` **also concluded `failure`**. The workflow still shows green because the job is `continue-on-error: true`, so the red never surfaces at workflow level.
- Reproduced locally: the same 14 tests fail on an unmodified `origin/main` checkout with `uv sync --frozen`.

The failures cluster into stale tests vs. refactored code, in files this PR does not touch:
- `test_notification_service.py` — asserts private helpers (`_build_subject`, `_build_html`, `_build_results_html`) that no longer exist on `NotificationService`, and raw-string recipients where fastapi-mail now yields `NameEmail` objects (same for `test_data_source_member_email.py`).
- `test_eval_draft_helpers.py` — patches `completion_feedback_service.LLMService`, which that module no longer imports.
- `test_notify_service.py`, `test_whatsapp_webhook_route.py`, `test_artifact_relationship_loading.py`, `test_mcp_tool_registry.py` — assertions against behavior that changed underneath them (e.g. the webhook handler now dies earlier on `'_Result' object has no attribute 'scalar_one_or_none'`).

No open PR carries a fix I could port. Fixing 14 unrelated unit tests would substantially widen this PR, so I'm leaving them for a dedicated cleanup; happy to take that on separately. A re-run won't change the outcome — the failures are deterministic.

---

PR #1028, opened 2026-08-27
read_file: show the document, not just its text

r0533130357-dev:
## What

`read_file` cards rendered a text excerpt even when the file was a PDF, spreadsheet or image. The result now carries an explicit `preview` contract and the card renders the real file.

| file | card shows |
|---|---|
| PDF (incl. `page_range`, scanned) | the PDF, opened at the page the model read |
| Excel / CSV | compact HTML table |
| PNG / JPEG | image, click to expand |
| Word / PowerPoint | text only — needs LibreOffice conversion, deferred |

**A document always beats its page renders.** A scanned PDF is rasterized so a vision model can read it, but the user is still shown the PDF at the right page.

Verified end-to-end against a live SharePoint connection: 9 of 10 real files preview correctly.

## Three defects fixed along the way

**`Content-Disposition` 500'd on non-Latin filenames.** HTTP headers are latin-1, and the filename was interpolated straight in, so any file named in Hebrew/Cyrillic/CJK returned `UnicodeEncodeError` on both the download and embed routes. Now emits both RFC 6266 forms, matching the pattern already used by the report/export routes.

**PDF text came out of `pypdf` in paint order**, which reverses every word of an RTL line. The characters are all valid, correctly-encoded Hebrew — just in the wrong order — so no garble heuristic could catch it and the model answered confidently wrong about the user's own document. Extraction moved to PDFium, already a dependency via `pypdfium2`. Identical output on Latin documents.

**`preview` was dropped by the persisted UI projection.** Report read endpoints serve `context_summary_json`, not `result_json`, so the preview rendered live and then vanished the moment a finished turn was refetched. Added to the allowlist.

Backfilled with a targeted script rather than bumping `CONTEXT_SUMMARY_VERSION`: that would invalidate `create_data`/`write_csv` projections too and reload their row payloads on **every** report open — measured at ~2.8 ms/MB, and never persisted, so the cost repeats forever.

## No second download

Connector clients already held the file bytes in memory when they extracted text and then discarded them. `DocumentText` carries them through (mirroring the existing `NamedBytes`), so the original document is stored for the viewer without re-fetching — confirmed against real SharePoint. `network_dir` keeps returning a plain `str`; re-reading a local file is free and holding every document in memory would not be.

## Notes

- `/embed` (inline) backs the PDF iframe; `/content` sets `attachment` and would download instead of render.
- Previews lazy-mount on scroll — a long conversation must not open dozens of iframes.
- New UI strings added to all 10 locales; layout uses logical properties for RTL.

## Verification

- 240+ backend tests; the new suites were confirmed to fail on `main` so they cover real defects
- Frontend build clean, code confirmed present in the shipped bundle
- Live checks against the running app and a real SharePoint tenant

Not yet exercised in a browser end-to-end.

### DIFF
--- a/backend/app/ai/tools/implementations/read_file.py
+++ b/backend/app/ai/tools/implementations/read_file.py
@@ -426,6 +426,12 @@ class ReadFileTool(Tool):
                 )
                 return
             shown = f"{paged.get('first')}-{paged.get('last')}"
+            # Bytes and name the client carried over from the fetch it already
+            # made (see the __doc_pages__ contract). Absent for network_dir,
+            # which extracts straight from disk — there the fallback re-read is
+            # local and free.
+            paged_raw = paged.get("raw")
+            paged_name = paged.get("name") or None
 
             # Scanned/image-only pages (no usable text), glyph-soup extractions
             # (text came back but it's garbled — subset font with a broken
@@ -445,10 +451,11 @@ class ReadFileTool(Tool):
             ):
                 import asyncio as _asyncio
 
-                raw_bytes = None
+                raw_bytes = paged_raw
                 try:
-                    raw = await _asyncio.to_thread(client.read_raw_bytes, data.file_id)
-                    raw_bytes = raw[0] if isinstance(raw, tuple) else raw
+                    if raw_bytes is None:
+                        raw = await _asyncio.to_thread(client.read_raw_bytes, data.file_id)
+                        raw_bytes = raw[0] if isinstance(raw, tuple) else raw
                     imgs, total = render_pdf_pages_images(raw_bytes, rng[0], rng[1])
                 except Exception as e:
                     imgs, total = [], paged.get("pages_total")
@@ -468,10 +475,12 @@ class ReadFileTool(Tool):
                         connection_id=data.connection_id,
                         session_file=session_file,
                         raw_bytes=raw_bytes,
+                        raw_name=paged_name,
                     )
                     output, observation = await self._finalize(
                         data, runtime_ctx,
-                        rendered={"content_type": "binary", "pages_shown": shown},
+                        rendered={"content_type": "binary", "pages_shown": shown,
+                                  "file_name": paged_name},
                         session_file_id=source_file_id,
                         image_pngs=[png for png, _mtype in imgs],
                         pages_total=total,
@@ -498,6 +507,8 @@ class ReadFileTool(Tool):
                 runtime_ctx, client, data.file_id,
                 connection_id=data.connection_id,
                 session_file=session_file,
+                raw_bytes=paged_raw,
+                raw_name=paged_name,
             )
             output, observation = await self._finalize(
                 data, runtime_ctx,
@@ -506,6 +517,7 @@ class ReadFileTool(Tool):
                     "text": paged.get("text") or "",
                     "pages_total": paged.get("pages_total"),
                     "pages_shown": shown,
+                    "file_name": paged_name,
                 },
                 session_file_id=source_file_id,
                 image_pngs=[],
--- a/backend/app/ai/tools/implementations/read_file.py
+++ b/backend/app/ai/tools/implementations/read_file.py
@@ -145,7 +145,8 @@ def _content_details(output: Dict[str, Any], *, max_chars: int) -> str:
 
 
 def _build_preview(output: Dict[str, Any], *, preview_file_id: Optional[str],
-                   first_image_mime: Optional[str] = None) -> Dict[str, Any]:
+                   first_image_mime: Optional[str] = None,
+                   source_is_image: bool = False) -> Dict[str, Any]:
     """The UI contract for this read (see ReadFilePreview).
 
     A DOCUMENT always beats its page renders: when a scanned PDF is rasterized
@@ -164,6 +165,11 @@ def _build_preview(output: Dict[str, Any], *, preview_file_id: Optional[str],
         "pages_total": output.get("pages_total"),
         "image_file_ids": None,
         "truncated": False,
+        # True when the images are page RENDERS of a document, not the source
+        # picture itself. The card auto-shows real pictures (they ARE the
+        # content) but keeps derived galleries — a lossy rasterized stand-in —
+        # collapsed behind a click.
+        "derived": False,
     }
 
     # page_range reports "10-15"; the viewer opens at the first of them.
@@ -195,6 +201,9 @@ def _build_preview(output: Dict[str, Any], *, preview_file_id: Optional[str],
             # Renders are capped (render_file_images max_pages), so a gallery
             # showing fewer pages than the document has must say so.
             "truncated": bool(pages_total and len(image_ids) < pages_total),
+            # A picture source stays "the picture" even when the id points at
+            # a normalized re-encode; only document rasters count as derived.
+            "derived": not source_is_image,
         })
         return preview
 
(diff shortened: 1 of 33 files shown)


backend/app/routes/file.py
```diff
@@ -125,6 +127,34 @@ async def remove_file_from_report(file_id: str, report_id: str, current_user: Us
...
+    parse, plus `filename*=UTF-8''…` carrying the real name for those that
+    understand it. Same shape the report / agent-export / step download routes
+    already use. Transliteration can strip a name down to nothing but its
+    extension (".pdf"), which browsers save as a hidden extensionless file — so
+    a stem with no letters or digits left is replaced outright.
+    """
+    name = filename or "file"
+    # A stray double quote would terminate the quoted-string early and corrupt
+    # the header, so it goes before anything else looks at the name.
+    ascii_name = name.encode("ascii", "ignore").decode("ascii").strip().replace('"', "")
```
yochze, line 154: **Header injection / recurring 500: control characters survive sanitization.**

`ascii_name` strips non-ASCII and double quotes, but ASCII control characters pass through untouched. A file named `'a\r\nX-Evil: 1.pdf'` (uploadable, or synced from a connector) produces:

```
inline; filename="a
X-Evil: 1.pdf"; filename*=...
```

h11/uvicorn rejects header values containing `\r\n`, so `GET /files/{id}/content` and `/embed` 500 again — the same failure class this fix targets. On servers that don't validate, the filename injects arbitrary response headers.

Fix: strip chars `< 0x20` and `0x7f` as well, e.g.

```python
ascii_name = "".join(
    c for c in name.encode("ascii", "ignore").decode("ascii")
    if c >= " " and c != "\x7f"
).strip().replace('"', "")
```

---

frontend/components/tools/ReadFileTool.vue
```diff
@@ -138,6 +191,22 @@ const previewText = computed(() => {
   return ''
 })
 
+// The server decides what to render — the stored session file is often a
+// derivative of the source (report.pdf -> report.pdf.txt), so the file name
+// and content_type here cannot answer it. Older tool executions predate the
+// contract and simply get no visual preview.
+const preview = computed<any>(() => rj.value.preview || { kind: 'none' })
+const showVisualPreview = computed(() =>
```
yochze: **Public share page regression: PDF previews can never load on `/c/[token]`.**

The share page (`frontend/pages/c/[token]/index.vue`) renders the same tool components, and the projection served there now carries `preview`, so `showVisualPreview` is true for PDF cards. But `FilePreview` → `useFileEmbedUrl` → `GET /api/files/{id}/embed_token` is guarded by `@requires_permission('manage_files')` and returns 401 for anonymous viewers; `getEmbedUrl` catches and returns `''`, leaving a permanent 260px "Preview unavailable" box on every `read_file` PDF card in a publicly shared report — a visible regression from the previous text-only card.

Suggest gating the visual preview per host page (like `canExpand`) or minting the embed token via the share token.

---

backend/app/ai/tools/implementations/read_file.py
```diff
@@ -380,71 +444,79 @@ async def run_stream(
                 and allow_llm_see_data(runtime_ctx)
             ):
                 import asyncio as _asyncio
-                import base64 as _b64
+
+                raw_bytes = None
                 try:
                     raw = await _asyncio.to_thread(client.read_raw_bytes, data.file_id)
                     raw_bytes = raw[0] if isinstance(raw, tuple) else raw
```
yochze: **Opaque-id page reads (SharePoint/Drive) persist the PDF but never show it.**

`raw` is a `(bytes, name)` tuple from `read_raw_bytes`, but only `raw[0]` is kept — the real filename is discarded. For an opaque Graph id, the page-range client dict carries no name and `_name_from_path_id` returns `''`, so `_finalize` leaves `output.file_name` unset and `_build_preview`'s `_doc_ext(file_name) == 'pdf'` gate fails: the vision branch yields `kind='none'`, the text branch `kind='text'`.

Meanwhile `persist_source_document` has already fetched and stored the full PDF (named by the opaque id, no extension, `mime=None`) — the cost is paid, but the headline feature (PDF opened at the read page) silently never appears for exactly the SharePoint case this PR targets.

The whole-file path fixed this via the connector-supplied payload name; the page-range branches need the same — keep `raw[1]` and thread it through as `file_name`.

---

backend/app/ai/tools/implementations/read_file.py
```diff
@@ -380,71 +444,79 @@ async def run_stream(
...
-            }
-            name = (
-                getattr(client, "display_name", None) if session_file is not None
-                else _name_from_path_id(data.file_id)
+            # Same deal for a text page read: the slice goes to the model, the
+            # whole document is kept so the card can open it at that page.
+            # Deliberately no cache write — the cache key carries no page
+            # dimension, so storing a slice under the whole-file key would
+            # later be served back as the entire file.
+            source_file_id = await persist_source_document(
```
yochze, line 519: **Text page-range reads double-download the whole remote file.**

This call passes neither `raw_bytes` nor `reuse_existing`, so `persist_source_document` → `read_source_bytes` → `client.read_raw_bytes` downloads the entire file again — after the Graph/S3 client already downloaded it in full to extract the page slice. Paging through a 20 MB SharePoint PDF (`page_range='1-5'`, then `'6-10'`, …) costs 2N full downloads for N page reads, violating the PR's own "no second download" rule.

At minimum `reuse_existing=True` (as the cache-hit path at line 579 does) skips the re-fetch for pages 2..N; threading the in-hand bytes through fixes the first read too.

---

yochze (commented):
Review of the visual-preview feature: 4 findings, all verified against the diff (2 blocking-worthy, 2 feature-completeness). Details inline.

1. `_content_disposition` still lets CR/LF (any ASCII control char) into the header — recurring 500 + header injection.
2. PDF previews on the public share page can never load (embed_token requires an authenticated org member) → permanent "Preview unavailable" boxes for anonymous viewers.
3. Page-range reads by opaque provider id (SharePoint/Drive) discard the real filename, so the PDF preview never appears despite the source being downloaded and persisted.
4. Text page-range reads re-download the entire remote file in `persist_source_document`.

Also checked and found fine: `manage_files` gating for members, backfill script's async-session usage, locale key coverage, CSV row-limit arithmetic, projection allowlist/cache-exclusion changes.

---

backend/app/routes/file.py
yochze: Fixed in 3c31f06: the source name is scrubbed of all non-printable characters (CR/LF, other C0 controls, and DEL) before either header form is built, and backslashes are dropped from the quoted-string so one can't escape the closing quote. Also `quote(name, safe='')` so `/` is percent-encoded in the RFC 5987 ext-value. Pinned by new cases in `test_content_disposition_header.py` (CR/LF injection, tab, path separators, trailing backslash).

---

frontend/components/tools/ReadFileTool.vue
yochze: Fixed in cc24e56: the share page already passes `readonly=true` to every tool component, and `ReadFileTool` now honors it — PDF and image previews (which authorize via `/embed_token` / `/content` and 401 for anonymous viewers) are suppressed on `/c/[token]`, restoring the pre-preview text-only card. The inline CSV table carries its own data in the projection, so it still renders for guests. Minting embed tokens via the share token remains possible as a follow-up if shared PDFs should render too.

---

backend/app/ai/tools/implementations/read_file.py
yochze: Addressed on the current head: the `__doc_pages__` contract now carries `name` alongside `raw` (graph_drive/s3 clients populate it precisely because opaque provider ids have no filename), and both page-range branches thread it through as `file_name` / `raw_name`, so `_build_preview`'s `pdf` gate resolves for SharePoint/Drive reads. Pinned by `test_carried_name_makes_the_preview_resolve` in `test_read_file_source_document.py` (passing).

---

backend/app/ai/tools/implementations/read_file.py
yochze: Addressed on the current head: the text page-range call now passes `raw_bytes=paged_raw` — the bytes the DocumentText client already fetched — so remote sources persist without a second download (`test_paging_costs_one_download_per_read` pins it; for `network_dir` the bytes are absent by design and the fallback re-read is local). On top of that, 3c31f06 makes repeated paged reads of an unchanged document skip the rewrite entirely: the refresh path compares size then bytes and reuses the existing row, avoiding the disk write, preview regeneration, and extra commits per page read.

---

yochze:
**CI note: the `unit-tests` failure on this PR is pre-existing on `main`, not this PR's.**

The failing check is `unit-tests` (13 failures, e.g. [run 33469291387](https://github.com/bagofwords1/bagofwords/actions/runs/33469291387/job/99735562105)). Evidence it's not from this branch:

- The **same job fails on `main`'s own latest push run** ([82dc78f, run 3068](https://github.com/bagofwords1/bagofwords/actions/runs/33345766959/job/99349462762)) — the workflow shows green overall only because `unit-tests` doesn't gate the dispatch.
- The failing tests are all in areas this PR doesn't touch: `test_notification_service` (missing `_build_subject`/`_build_html`), `test_notify_service`, `test_data_source_member_email` (`NameEmail` vs plain string — looks like fastapi-mail drift), `test_eval_draft_helpers` (`LLMService` attribute moved), `test_mcp_tool_registry`, `test_whatsapp_webhook_route`.
- I also reproduced the identical set locally at this PR's fork point (`5d1ea14`) in a clean worktree.

No fix exists on `main` yet to port. Everything this PR touches is green: the read_file/preview suites pass locally (67 tests incl. the new ones), and `integration-llms`, `integration-data-sources` are green on the current head (`e57d425`); e2e/playwright still running. I'll keep watching, and if a fix for these lands on `main` I'll bring it into this branch.

---

PR #1030, opened 2026-08-28
docs(design): diagnose artifact iteration latency, edit regressions, and dashboard filtering gaps

yochze:
Customer-reported pain: iterating an artifact is slow, edits regress other
areas of the page, and dashboards lack Tableau-style filters/parameters.

Documents the root causes with file:line evidence (per-edit Chromium
render-repair loop, >30% rewrite routing, positional viz[N] indexing,
history-free edit prompts, params opt-in only at create_data time) and a
phased remediation plan (S/R/F items) with sequencing.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013aKhfXooEvD8NWCuCA3hTd


yochze:
Review — changes requested

The direction is good: free-form artifact source plus stable data IDs and edit-first iteration is the right architecture. I would not merge this version yet because the main reliability guarantees are not actually enforced.

Merge blockers:

1. The visualization reference gate is advisory. After repair attempts, create_artifact marks the candidate clean whenever rendering is non-fatal, even when reference or wiring errors remain, so invalid artifacts are still persisted.
2. The coverage checker has large false-negative paths. A nonempty visualization payload with no references passes; length, filter, and slice are all treated as complete coverage. I verified that a chartless page and slice(0, 1) both return no errors.
3. The filter fix does not reach the active production planner. V3 still directs client-side derivation and does not require server-side parameters across every affected visualization. The existing validator only searches source text and never proves that each visualization participates, so a filter can still affect only a subset and pass.

Additional reliability issues:

- Large edits are routed to edit_artifact, but its coder prompt says to output nothing when diffs are too large and promises the planner will switch to create_artifact. The new full-rewrite fallback only works if the coder disobeys that instruction.
- The legacy positional codemod rewrites raw source, including strings and comments. A valid string containing viz[0] can become invalid nested-quote JavaScript.
- The pinned-decisions digest retains superseded decisions as coequal requirements, such as dark mode followed by light mode.
- Rebuild preservation includes only the first 30,000 characters while instructing the model to reproduce everything, so later sections of large dashboards remain vulnerable to deletion.

The vizById() API is the strongest part and matches the desired flexible data-object approach. I would extend that boundary model to host-owned filter and parameter bindings while leaving JSX and layout unconstrained.

Validation performed: 32 focused backend tests passed, artifact-globals.js passed syntax checking, and git diff --check was clean. The tests currently encode the empty-code coverage blind spot, and I found no E2E evidence supporting the design document claim that phases 1–2 were verified E2E.

No code changes were made.

PR #1039, opened 2026-08-31
read_file previews: fix thread-safety, attachment attribution, and preview edge cases

yochze:
Follow-up to #1028, targeting its branch: eight fixes for issues found in review, each verified against a live sandbox (real backend + scripted LLM driving `read_file` through the chat UI).

## Blockers

**1. PDFium is not thread-safe — all pypdfium2 use now takes a shared lock.**
Moving text extraction to PDFium routed every PDF read/search/grep through a native library documented as not thread-safe, on `asyncio.to_thread` workers, concurrently with the existing rasterization paths. Two simultaneous PDF reads were a native data race that can corrupt PDFium state or segfault the backend. `PDFIUM_LOCK` (module-level, in `_document_text.py`) now guards every open→read→close sequence: extraction, `render_pdf_pages_images`, and the PDF branch of `render_file_images`.

**2. Tool-generated page renders showed up as attachments on the next user message.**
`attach_drive_file_to_session` links renders to `report.files` with a NULL `completion_id`, and `_mark_images_with_completion` treats every unclaimed image in the report as a pending upload — so scanned-PDF page renders got stamped onto the *next* unrelated user message and rendered as attachments the user never sent. The sweep is now scoped to `source_kind="upload"` (non-null with a server default, so legacy uploads still match). Verified in the sandbox: `scanned_invoice.pdf.p1/p2.png` stay unclaimed across five later messages.

**3. Paged reads advertised the raw source PDF as `session_file_id`.**
The output schema tells the model to pass `session_file_id` to `inspect_data`/`read_excel_as_csv` — a guaranteed parse failure on a raw PDF. The viewer copy now reaches the UI via `preview.file_id` only; it is attached with `analysis=False` (kept off the live roster), and `#source` rows are excluded from `analysis_files` at agent init so they can't rejoin next turn.

## Fixes

**4. `_content_disposition` hardening.** `quote(name, safe='')` so `/` is percent-encoded in the RFC 5987 ext-value (connector names are often full paths; strict parsers drop the whole `filename*` otherwise); control characters (CR/LF) are stripped before either form is built (they survive an ascii encode and 500 at header send time); backslashes are removed so one can't escape the closing quote.

**5. Paged reads no longer rewrite the whole document per read.** The refresh-in-place path now compares sizes (free) and then bytes, and returns the existing row untouched when identical — skipping the full disk rewrite, the preview regeneration, and several commits per `page_range` read. Verified live: the second paged read of the same PDF logs `unchanged, reusing session file` instead of rewriting.

**6. Embed tokens re-mint before expiry.** A mounted iframe kept its original token URL forever (`visible` never flips back, so the watch never refired), so a tab open past the 1-hour TTL showed a dead 403 frame. `FilePreview` now schedules a refresh keyed to the cached token's age (`msUntilRefresh`) and the iframe src swaps reactively; the timer is cleared on unmount.

**7. TIFF is no longer offered as a browser-renderable image.** `is_picture_name` now tests against `_BROWSER_IMAGE_EXTS` (renderable minus tiff/tif), so a TIFF read falls back to its PNG renders instead of handing the UI a broken ``. The vision path keeps decoding TIFF via PIL as before.

**8. Side-panel image click works.** The panel's `FilePreview` instance had no `@open` listener, so the advertised zoom click did nothing. It now opens `ImagePreviewModal`, same as the in-card instance. Verified interactively in the sandbox.

## Tests & verification

- Three tests in `test_read_file_source_document.py` updated to the new paged-read contract (doc reaches the UI via `preview.file_id`, not `session_file_id`; viewer copies attach with `analysis=False`); new cases for the header edge cases (path separators, CR/LF, trailing backslash) and the tiff split.
- Full backend unit suite on this branch: **3788 passed, 14 failed, 9 skipped** — and the same 14 tests fail identically on the unmodified PR head (`5d1ea14`, verified in a separate worktree), all in areas this diff does not touch (notification/email services, eval draft helpers, MCP tool registry, Sybase source, WhatsApp webhook, artifact relationship loading). No regressions introduced.
- Frontend production build clean.
- Live sandbox end-to-end on this branch: scanned/paged/RTL PDFs, CSV, XLSX, JSON, PNG, DOCX all preview correctly from both the live stream and the persisted `context_summary_json` projection; no attachment thumbnails on subsequent user messages; paged reads carry no `session_file_id` and reuse the stored document.

https://claude.ai/code/session_01MZd1zWJdkvwU4yHQBv18CR

### DIFF
--- a/backend/app/ai/agent_v2.py
+++ b/backend/app/ai/agent_v2.py
@@ -586,7 +586,15 @@ class AgentV2:
                 if (getattr(f, 'content_type', '') or '').startswith('image/')
                 and not _is_connector(f)
             ]
-            self.analysis_files = [f for f in all_files if not (getattr(f, 'content_type', '') or '').startswith('image/')]
+            # "#source" rows are persist_source_document's viewer copies of
+            # original documents (the .pdf behind a .pdf.txt session file).
+            # They are linked to the report for the embed route, but they are
+            # not analyzable and must not rejoin the roster at init.
+            self.analysis_files = [
+                f for f in all_files
+                if not (getattr(f, 'content_type', '') or '').startswith('image/')
+                and not str(getattr(f, 'source_ref', '') or '').endswith('#source')
+            ]
         else:
             self.data_sources = []
             self.clients = {}
--- a/backend/app/ai/tools/implementations/_file_tool_common.py
+++ b/backend/app/ai/tools/implementations/_file_tool_common.py
@@ -426,26 +426,31 @@ def render_pdf_pages_images(pdf_bytes: bytes, first: int, last: int, *, max_page
     Raises on an unreadable PDF."""
     import pypdfium2 as pdfium
     from app.ai.llm.image_utils import encode_pil_for_vision
+    from app.data_sources.clients._document_text import PDFIUM_LOCK
 
-    pdf = pdfium.PdfDocument(bytes(pdf_bytes))
-    try:
-        total = len(pdf)
-        first = max(1, int(first))
-        last = min(total, int(last))
-        out = []
-        for i in range(first - 1, last):
-            if len(out) >= max_pages:
-                break
-            page = pdf[i]
-            try:
-                bitmap = page.render(scale=dpi / 72.0)
-                pil = bitmap.to_pil()
-                out.append(encode_pil_for_vision(pil))
-            finally:
-                page.close()
-        return out, total
-    finally:
-        pdf.close()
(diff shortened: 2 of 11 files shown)


yochze:
Closing: per the author's request these fixes were pushed directly onto `feat/read-file-visual-preview` (commit 3c31f06, a fast-forward of this PR's head), so they are now part of #1028 and this PR's diff is empty. All findings, fixes and verification details are preserved in this PR's description and in the commit message.

---

PR #1051, opened 2026-09-03
Add N-prefix for non-ASCII strings in T-SQL clients

yochze:
## Summary
This change adds support for prefixing non-ASCII string literals with the `N` prefix in T-SQL clients (MSSQL and MS Fabric). This ensures that non-ASCII characters are properly handled as Unicode (nvarchar) rather than being silently replaced with `?` characters when the database uses a limited collation like SQL_Latin1_General.

## Key Changes
- Added `_UNICODE_PREFIX_CLIENTS` set containing `MSSQLClient` and `MsFabricClient` to identify T-SQL dialects
- Modified `_render_literal()` function to accept a `unicode_prefix` parameter that adds the `N` prefix to non-ASCII string literals
- Updated `render_sql_with_params()` to detect T-SQL clients and pass the `unicode_prefix` flag through the rendering pipeline
- ASCII-only strings remain unprefixed to preserve type and index behavior for varchar column comparisons
- Non-ASCII characters in list items are individually prefixed as needed
- Quote escaping behavior is preserved for both ASCII and non-ASCII strings

## Implementation Details
- The `N` prefix is only applied when both `unicode_prefix=True` and the string contains non-ASCII characters (checked via `str.isascii()`)
- The solution handles nested structures like lists by recursively applying the same logic to each item
- Other SQL dialects (SQLite, PostgreSQL, MySQL) are unaffected and continue to render strings without the prefix
- Comprehensive test coverage includes: basic non-ASCII handling, ASCII passthrough, list handling, quote escaping, and cross-client verification

https://claude.ai/code/session_01Xx9FC1hBVR3U1UGDYWdWHi

### DIFF
--- a/backend/app/ai/code_execution/query_params.py
+++ b/backend/app/ai/code_execution/query_params.py
@@ -196,10 +196,20 @@ def resolve_param_values(
 # default; both ' and \ must be escaped there.
 _BACKSLASH_DIALECT_CLIENTS = {"MysqlClient", "MariadbClient", "MariaDBClient"}
 
+# T-SQL clients: a plain 'literal' is a varchar in the database's default
+# collation, so any character outside that code page is silently replaced by
+# '?' at parse time (a Hebrew value on a SQL_Latin1_General database becomes
+# '???' and matches nothing). Non-ASCII strings must be N'literal' (nvarchar)
+# there. ASCII stays plain so comparisons against varchar columns keep their
+# type and index behaviour.
+_UNICODE_PREFIX_CLIENTS = {"MSSQLClient", "MsFabricClient"}
+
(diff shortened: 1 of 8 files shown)


yochze:
CI status on this PR:

- **integration-data-sources** fails on `tests/integrations/ds_clients.py::test_data_source_connection[snowflake]` with `000666 (57014): Your account is suspended due to lack of payment method.` from Snowflake. It reproduced identically on both heads (`5405f79`, `40bdceb`). The diff does not touch the Snowflake client or the integration harness; this is the shared test account's billing state and needs fixing on the Snowflake side, not in the repo. No code fix exists to port.
- **unit-tests** was already red on `main` (its own latest run, `033e223`, fails the same 13 tests). `40bdceb` repairs those stale tests; none of them concerned query params.

---

PR #1054, opened 2026-09-03
Add side panel view for query results with full-width preview

yochze:
## Summary

Query results from tools like `create_data`, `read_query`, `write_csv`, and `describe_entity` can now be opened in the report's side panel at full width and height, similar to how documents open from `read_file` cards. This provides a better viewing experience for charts and tables that were previously constrained to a narrow inline card format.

## Key Changes

- **ToolWidgetPreview component** (`frontend/components/tools/ToolWidgetPreview.vue`):
  - Added `canExpand` and `expanded` props to support side panel rendering
  - Added "Open in side panel" button (when `canExpand` is true) that emits `openDataPanel` event
  - Converted fixed heights (`h-[340px]`, `h-[400px]`, `h-[250px]`) to flexible layouts when `expanded=true`
  - Disabled collapse toggle when in expanded mode (header stays pinned as toolbar)
  - Fixed window event listener management to use per-instance listeners instead of global slots (prevents listener conflicts when multiple previews are mounted)
  - Updated row count display to report `total_rows` when data is truncated, with a spinner during hydration
  - Added `data-testid` attributes for testing

- **Report page** (`frontend/pages/reports/[id]/index.vue`):
  - Added `'data'` to `rightPanelView` union type
  - Implemented `panelData` state and `openDataPanel`/`closeDataPanel` functions to manage the data panel lifecycle
  - Added transient data panel tab with icon, title, and close button
  - Renders `ToolWidgetPreview` in expanded mode within the side panel
  - Panel returns to the previous view (dashboard, agent, summary, etc.) when closed

- **Tool components** (CreateDataTool, CreateWidgetTool, ReadQueryTool, DescribeEntityTool, WriteCsvTool, ChatSummary):
  - Forward `canExpand` prop to ToolWidgetPreview
  - Re-emit `openDataPanel` event to parent

- **Localization** (`locales/{en,es,he}.json`):
  - Added strings: `openInPanel`, `closeDataPanel`, `dataTab`, `loadingRows`

- **Setup script** (`tools/agent/setup_haiku_llm.py`):
  - Fixed model default setting to use dedicated `set_default` endpoint instead of PATCH (which doesn't support `is_default` field)

- **Testing & Documentation**:
  - Added `tests/reports/shoot-data-side-pane.mjs` — end-to-end test that drives a real LLM prompt, opens the result in the side panel, and validates all controls and geometry
  - Added `docs/feedback-loops/toolwidget-preview-side-pane.md` — detailed feedback loop documentation
  - Added screenshot evidence in `media/pr/toolwidget-preview-side-pane/`

- The side panel reuses the exact same `ToolWidgetPreview` component as the inline card, ensuring feature parity and reducing maintenance burden
- Window event listeners are now tracked per-instance to prevent conflicts when multiple previews mount/unmount
- Row count label now correctly reports the server's total row count during truncated preview states, with a loading spinner
- Desktop-only feature (`:can-expand="!isMobile"`) to maintain mobile UX
- Panel state is transient and returns to the previous view when closed, matching the file preview pattern

https://claude.ai/code/session_013hiHVCT4BoS5vCkhdEPa9E


yochze:
CI status note: the `integration-data-sources` check is red on this PR, but the failure is not this change's.

- Failing test: `tests/integrations/ds_clients.py::test_data_source_connection[snowflake]`
- Error: `snowflake.connector.errors.ProgrammingError 000666 (57014): Your account is suspended due to lack of payment method.`
- The same job fails identically on `main` at the base commit 95e83b9 ([run 33784387922](https://github.com/bagofwords1/bagofwords/actions/runs/33784387922/job/100745519856)), two hours before this PR existed.
- This PR touches only frontend components, locale catalogs, an agent helper script, docs and evidence images; it does not touch the Snowflake client or the integration tests.

There is no code fix to port: the Snowflake test account needs its billing restored. A re-run will not change the outcome until then, so I have not re-run it. Every other check on this head is passing or still running; I will keep watching the PR.

---

PR #1061, opened 2026-09-04
Move queries from standalone page into agents knowledge tree

yochze:
## Summary
Queries have been migrated from their own standalone page (`/queries`) into the agents knowledge tree in the `/agents` explorer. Each agent now has a "Queries" row that opens a panel listing its saved queries, with full detail view support and proper visibility controls.

### Frontend
- **New Components**:
  - `EntityDetailPanel.vue`: Extracted detail view (header, tabs, parameters, actions) previously embedded in `/queries/[id]`
  - `AgentQueriesPanel.vue`: New list panel for queries under an agent in the tree

- **Modified Components**:
  - `KnowledgeExplorer.vue`: Added Queries row under each agent; manages panel state and query counts
  - `/pages/queries/[id]/index.vue`: Now delegates to `EntityDetailPanel` component
  - `/pages/queries/index.vue`: Simplified to redirect or show legacy content

- **Navigation**: Queries are now accessed through the agent tree rather than a top-level page; deep-linking to `/queries/[id]` still works via the detail page

### Backend
- **New Module**: `app/core/entity_scope.py` - Implements workshop visibility rules (who can see unpublished entities)
  - Catalog tier: published entities visible to anyone who can reach all their agents
  - Workshop tier: drafts/suggestions/archived visible only to owner, org admins, or managers with `create_entities` on all agents

- **API Changes**:
  - `GET /api/entities/counts`: New endpoint for per-agent query counts in tree badges
  - `GET /api/entities`: Enhanced with visibility filtering via `entity_authority_scope()` and `visibility_clause()`

- **Tests**: Added `test_entity_workshop_visibility.py` to verify visibility rules at the API level

### Localization
- Added translations for "Queries", "No queries yet", and "Untitled query" in en, es, and he locales

### Testing
- Updated test descriptions to reflect new query location in the tree
- Visibility tests now cover the API-level workshop rules

## Implementation Details
- Visibility filtering is applied in SQL before pagination, preventing workshop rows from consuming page slots
- Per-agent badge counts use the same visibility rules as the list, ensuring consistency
- The detail view is reusable: renders over an agent's query list (with back button) or standalone for deep-linked queries
- Backward compatibility maintained for direct `/queries/[id]` links

https://claude.ai/code/session_01SYgUXSgxzGkaVcynqMBzTU

PR #1060, opened 2026-09-04
Add PDF export for document artifacts via headless rendering

yochze:
## Summary

Implements PDF export for document artifacts by rendering the frontend's standalone paper page through headless Chromium, rather than attempting server-side markdown rendering. Documents now export as properly formatted, multi-page PDFs instead of being refused or printed as narrow ribbons via the browser's print dialog.

## Key Changes

**Backend (PDF Service)**
- Added `generate_doc_pdf()` and `_generate_doc_pdf_inner()` methods to render documents through `/print/doc` page with injected data
- Extracted common render timeout/cancellation logic into `_bounded_render()` shared by dashboard and document renderers
- Added helper functions: `_doc_page_geometry()` (portrait layout at document margins), `_chromium_executable()` (deployment-specific browser path), `_render_origin()` (headless browser's internal URL)
- Added regex pattern `_FILE_PLACEHOLDER_RE` to extract embedded file IDs from markdown (images must be inlined as data URIs since headless renderer has no session)
- Added JavaScript helpers: `_DOC_CHARTS_READY_JS` (waits for ECharts to render), `_DOC_TABLE_BREAKS_JS` (allows tables to break across pages with repeated headers, unlike charts)
- Updated `generate_for_report()` to include "doc" mode in artifact preference order (dashboard > deck > document)
- Updated `_render_artifact_pdf()` to call `_render_doc_pdf()` for document artifacts instead of refusing them

**Frontend (Paper Page)**
- Created `frontend/pages/print/doc.vue`: standalone page that receives document data via `window.__BOW_DOC__` injected script, renders DocViewer in paper mode, and signals readiness to headless renderer
- Added comprehensive print stylesheet with document-appropriate typography (points, not pixels), margins matching Word defaults, and page break rules
- Updated `DocViewer.vue` to accept `paper` and `files` props for headless rendering (files as pre-resolved data URIs)
- Updated `DocVizEmbed.vue` to render tables as full HTML tables on paper (no virtualization) instead of AG Grid, with proper formatting and numeric column alignment
- Updated `DocEditor.vue` to track unsaved changes and emit `export-pdf` event to parent; PDF export now delegates to parent rather than using browser print
- Updated `ArtifactFrame.vue` to handle doc PDF export through shared ExportMenu (removed print button from read-only viewer)
- Updated `DocFileEmbed.vue` to accept pre-resolved `dataUri` for headless renders

**Routes & Configuration**
[...]

PR #1058, opened 2026-09-03
Umbrella 0.0.554: side panel, scikit-learn, VMware Aria, ServiceNow attachments, Data modal, file search

yochze:
Umbrella PR folding in the open PRs opened in the last 24 hours, plus the `VERSION` bump to 0.0.554 and one changelog line per PR. Each branch is merged into this branch, so merging this PR lands all of them and closes them.

Merged branches (latest tips):
- #1054 — Side panel view for query results with full-width preview
- #1055 — scikit-learn support with ML training constraints
- #1056 — VMware Aria Operations connector for infrastructure RCA
- #1057 — ServiceNow connector can fetch record attachments as files
- #1052 — Data modal to manage a dashboard's queries from the artifact panel
- #1059 — Find files by their containing folder; calmer, more compact read_file PDF card

Also included:
- Removed stray, unresolved merge-conflict markers left in `frontend/public/mcp-artifact-app.html` on `main` (introduced by commit 2752082, "Re-merge PR #1033"). The `theirs` side was empty, so the fix keeps the ARTIFACT_GLOBALS_VERSION sync comment and drops the markers.

Conflict resolution:
- `CHANGELOG.md` / `VERSION` — kept the concise umbrella entry (one line per PR at 0.0.554).
- `frontend/components/tools/ToolWidgetPreview.vue` (#1052 × #1054) — unioned the added props (`canExpand`/`expanded` from #1054, `canEdit` from #1052); both features' template bindings retained.
- `frontend/components/FilePreview.vue` (#1059 × #1054) — `frameHeight` now sizes only the inline card (`200px`, #1059's compact intent); the expanded side panel stays flex-driven per #1054's template (`:style="expanded ? undefined : { height: frameHeight }"`), so #1059's `70vh` expanded branch was dropped as dead code.

CI note: `integration-data-sources` fails on the Snowflake connection test with "account suspended due to lack of payment method" — a base-branch billing issue unrelated to this change, red on `main` as well.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018je4Dty9qr2yQjFyR7jZ9p

PR #1067, opened 2026-09-05
Umbrella 0.0.555: 15 PRs — doc PDF export, knowledge-tree queries, MSSQL Kerberos, session revocation, SSO auto-start, Prometheus, CloudWatch, Astra, skills catalog, FK introspection, SharePoint Server, tables ERD, onboarding picker, member deactivation, auth page refresh

yochze:
Umbrella PR folding in the open PRs opened in the last 24 hours, plus one changelog line per PR at 0.0.555. Each branch is merged into this branch, so merging this PR lands all of them and closes them. All fifteen head SHAs are verified ancestors of this branch, re-checked with `git ls-remote` against the live refs rather than a cached listing.

Merged branches (latest tips):
- #1060 — PDF export for document artifacts, rendered through the real document viewer in headless Chromium
- #1061 — Queries moved from their own page into the agents knowledge tree, with workshop visibility rules
- #1062 — Hardened SQL Server Kerberos delegation (process-global ccache lock, per-worker cache namespace)
- #1063 — Revocable session tokens, and the session JWT out of the SSO redirect URL
- #1064 — Single-provider SSO auto-start and `login_hint` passthrough
- #1065 — Prometheus connector out of `dev_only`, shipping as beta
- #1066 — AWS CloudWatch connector (log groups via Logs Insights + metrics on one connection)
- #1068 — GPT-6 Astra added to the OpenAI catalog, with GPT-6 routed through the Responses API
- #1069 — Foreign key introspection for SQL data sources, with FK pruning in schema context
- #1070 — A catalog of twelve pre-built skills an admin enables per organization (merged at `8c5ca0c`)
- #1071 — SharePoint Server (on-premises) file connector over NTLM/Kerberos
- #1072 — Shared table selection ERD canvas, List/Visual views (merged at `f957e54`)
- #1073 — Onboarding connector step leads with popular sources plus full-catalogue search
- #1074 — Removing a member's last organization membership deactivates the account instead of leaving a working login
- #1075 — Unified `AuthProviderButtons` component and refreshed sign-in/sign-up pages, with providers named and branded from their issuer URL

`VERSION` was already moved to 0.0.555 on #1061, so this PR only adds the release notes.

**Two things a merger should know:**
- #1063 signs every user out once on deploy, by design — existing tokens predate the revocation check.
- #1074 changes what "remove member" means on single-organization installs: the account is deactivated, not deleted. Content and history are kept and re-inviting restores access.

## A deliberate change to #1068 as merged

#1068 made GPT-6 Astra the OpenAI **default**, replacing GPT-5.6 Terra. Astra is roughly 4x Terra's input price and 3.3x its output price ($10/$50 vs $2.50/$15 per million), so that would move every organization onto a pricier model on upgrade without anyone choosing it. This umbrella keeps **Terra as the default** and ships Astra as a selectable model (`a2df9db`); Luna remains the small-model default. Everything else #1068 fixed is untouched — GPT-6 requests omit `temperature` on all three inference paths, honor reasoning settings, and route tool calls through the Responses API even behind a custom base URL. `docs/feedback-loops/astra-model.md` was updated to match what ships.

## Regressions caught in the merge, and fixed here

#1075 branched from `main`, so it carried none of #1063/#1064/#1074 — and it rewrites `frontend/pages/users/sign-in.vue`, the one file all three changed. A clean merge would have shipped two regressions:

- **#1064's `login_hint` passthrough was silently dropped.** The extracted `AuthProviderButtons` component calls `/api/auth/<name>/authorize` with no query at all, so an embedding app that passes a hint would have been sent back to a provider chooser. Fixed by adding a `loginHint` prop to the component, forwarding it on the authorize call, and wiring it from `sign-in.vue` — both the button path and the auto-start path now carry the hint.
- **The conflict hunk deleted `signInWithProvider`/`signInWithGoogle`, which #1064's single-provider auto-start still calls.** They are kept, with a comment explaining why they cannot delegate to the new component: the auto-start runs in `onMounted`, before `pageLoaded` renders it, so a template ref is still null at that point.

Independently of the merge, #1075 deleted `data.setupTitle` and `triggers.setupTitle` from all ten locales while `IntegrationConnectionForm.vue` and `automations/TriggersTab.vue` still reference them — twenty translations, and two labels that would have rendered as raw key paths. All twenty are restored from `main` in their original positions.

All fifteen behaviours from #1063 (login-code exchange), #1064 (auto-start, hint resolution including the nested-`redirect` case) and #1074 (typed `AppError`, all-caps sentinel fallback) were re-verified present on the merged tree, and `tests/unit/test_sso_login_hint.py` — #1064's own coverage for the restored feature — passes. All three fixes are re-checked after every subsequent re-merge.

## Two changes made here, beyond merging

- **CloudWatch is no longer `dev_only`.** #1066 shipped `aws_cloudwatch` gated behind `dev_only=True`, which hides it from the catalog outside a dev environment — so the connector it adds would not have been reachable by anyone on a normal deployment. It now ships visible as `version="beta"`, matching how #1065 shipped Prometheus in the same release. Verified inert for the `integration-data-sources` job, which drives a hardcoded `DATA_SOURCES` list that does not include cloudwatch (identical `1 failed, 3 passed, 15 skipped` before and after). `timbr_a2a` is now the only remaining `dev_only` connector, deliberately.
- **The README connector table and the Mintlify docs were brought up to date.** The new connectors from this release were missing, and so were eight older ones that had drifted out (including VMware Aria Operations). The table was regenerated against `data_source_registry.py` and checked programmatically, so it now covers every user-visible connector rather than a hand-maintained subset. SharePoint Server reuses the SharePoint icon in the app, the README and the docs. Docs changes are in bagofwords1/bagofdocs#22, which is separate from this PR and still open.

## A fix carried here that #1072 still needs

#1072 added `frontend/tests/tables_canvas/tables-canvas.spec.ts`, which read a locally-seeded fixture at **module scope**:

```ts
const seed = JSON.parse(fs.readFileSync(process.env.BOW_ERD_SEED || '/tmp/bow-tables-erd-run/seed.json', 'utf8'))
```

That file is produced by its own `tools/agent/seed_tables_canvas.py` on a developer's machine. CI has none, so the read threw during collection — and because it threw at import rather than inside a test, Playwright aborted the whole run:

```
before:  Error: ENOENT ...seed.json  ->  Total: 0 tests in 0 files
after:                                   Total: 86 tests in 21 files
```

So it was not one failing test: the entire Playwright suite stopped running. `playwright-tests` is red on #1072's own branch for this reason. The spec has now been re-merged six times and none of those commits changed that line. Here the seed is optional — its absence skips only that file, with a message on how to produce it, and the ERD tests still run locally via `BOW_ERD_SEED`. Wiring that spec to seed itself belongs in #1072.

## Conflict resolution

- `CHANGELOG.md` (#1061 × #1063 × #1065 × #1070 × #1072 × #1073 × #1074 × #1075) — one 0.0.555 section, one line per PR, replacing the `## Unreleased` headings the branches carried. #1070's entry has been rewritten on each of its six updates as the catalog went sixteen → ten → eleven → twelve skills; the count is cross-checked against the skill library directory each time. #1074's four bullets were folded into one line, and #1075 was merged without a changelog entry at all — caught by counting distinct `(#10xx)` refs against the merged branch list, and added.
- `frontend/components/KnowledgeExplorer.vue` (#1061 × #1070) — both add a panel to the same pane and the same `panelView` union. Resolved as a union, not a pick: #1061's `AgentQueriesPanel` keeps the leading `v-if`, #1070's `InstructionsSkillCatalogPanel` becomes the next `v-else-if` (carrying its `@open-instruction="openInstructionById"` binding) so the `evals`/`settings` chain below still reaches, and both entries go into the `kind` union and the `panelKindLabel` map. `skillCatalog.title` was deleted in #1070's rework; confirmed nothing still references it, and all 26 i18n keys the reworked panel uses resolve.
- `frontend/pages/users/sign-in.vue` (#1063 × #1064 × #1074 × #1075) — see **Regressions caught in the merge** above.
- `locales/*.json` (ten files, #1070 × #1072 × #1075) — three-way merged from the git stages. The conflicts were adjacency only: #1072's `tableErd` block sits beside #1070's reworked `skillCatalog` block, and no key changed on both sides.

## Overlaps that merged clean but were checked anyway

- **`backend/app/core/auth.py` (#1063 × #1064 × #1074 × #1075)** — four PRs touch this file. `user_lifecycle`'s docstring claims all three authentication paths re-assert `is_active` per request, which is what makes deactivation close an existing session rather than just future logins. That claim was verified on the merged tree rather than assumed: the JWT dependency is still `fapi.current_user(active=True)`, OAuth still filters `User.is_active.is_(True)` in `_active_subject`, and `ApiKeyService.get_user_by_api_key` now rejects an inactive user *before* bumping `last_used_at` (service-account keys exempt, since their backing row is inactive by construction).
- **`backend/app/services/auth_providers.py` (#1063 × #1064 × #1075)** — #1075 refactors `_is_entra_provider`/`_is_google_provider` onto the new `brand_for_issuer()`. #1063's `login_code` handler still runs before #1064's auto-start guard and returns on every path, so a callback arriving with `?login_code=` can't be bounced back out to the provider — no redirect loop. #1064's per-request `login_hint` still spreads *after* `extra_authorize_params`, so it beats the static default as intended.
- **`backend/app/schemas/data_source_registry.py` (#1062 × #1065 × #1066)** — checked per entry, since `dev_only=True` and `client_path=None` recur across entries: `prometheus` and `aws_cloudwatch` both have `dev_only` removed and `version="beta"`, and MSSQL carries its explicit `client_path`.
- **`frontend/components/datasources/TablesSelector.vue` (#1069 × #1072)** — #1069 adds foreign-key relationship badges to the list rows; #1072 adds the ERD canvas beside it. Both survive, re-verified after each of the six times #1072 was re-merged.
- **`backend/uv.lock` (#1069 × #1071)** — both add a dependency (`requests-gssapi`, the `mssql` testcontainers extra). The lockfile auto-merged; rather than trust that, it was validated with `uv lock --check`, resolving clean at 260 packages.
- **`backend/app/data_sources/clients/mssql_client.py`** — `c471b14` (MSSQL engine initialization under Kerberos cache activation) was pushed onto this branch while the #1070/#1072/#1074 merges were in flight. It was merged in rather than overwritten, and its new unit test passes on the merged tree.

## Verification

Run on this branch's head, not on the individual branches:

| Check | Result |
|---|---|
| `yarn build` (full frontend) | exit 0, fresh chunk hashes confirming it compiled this tree |
| Built bundle carries the restored fixes | `login_hint` in 3 chunks, `Set up trigger` in 1 |
| Playwright collection | 86 tests in 21 files (was 0 before the seed fix) |
| Frontend `node --test` suites | 9 passed |
| `tests/e2e/test_membership.py` + `test_scim.py` (#1074) | 39 passed |
| #1070 skill-catalog e2e + RBAC admin gate | 17 passed |
| Auth unit suites incl. `test_sso_login_hint.py` (#1064) | 83 passed |
| `tests/unit` incl. #1062's `test_mssql_kerberos.py` | 34 passed |
| All 12 skill files carry complete frontmatter | confirmed; count matches the changelog |
| `alembic heads` | one head (`sessepoch01`), no branch |
| Backend byte-compile (3.12) | clean |
| Conflict-marker scan | clean (the `SEARCH`/`REPLACE` tokens in `edit_artifact_legacy.py` are that tool's own diff format, in a file byte-identical to `main`) |
| All 15 PR heads are ancestors of this branch | confirmed via `git ls-remote` |
| `data.setupTitle` / `triggers.setupTitle` resolve in all 10 locales | confirmed |

**Environmental failures, not code failures:**
- `tests/e2e/test_llm_providers.py` — 5 tests fail here with `OPENAI_API_KEY_TEST is not set`, aborting in fixture setup before any test body runs. The fixture (`tests/fixtures/llm.py`) is unchanged from `main` by this branch.
- #1071's `sharepoint_onprem_client.py:149` sets `session.verify = os.environ.get("REQUESTS_CA_BUNDLE") or True`, and its tests assert `session.verify is True`. That holds only where `REQUESTS_CA_BUNDLE` is unset — true on CI, where #1071 is green, but not in a sandbox behind a proxy that sets it. Worth tightening in #1071 itself; nothing here is red on CI for it.

`integration-data-sources` is red for a reason established in the comments below (a Snowflake trial account suspended for lack of a payment method, errno 000666) and unrelated to this branch.

https://claude.ai/code/session_01AnjP3gwDW1mZRhmcbWMyhj


yochze:
`integration-data-sources` is red on this PR, and it isn't this PR's failure.

The job has exactly one failing test:

```
FAILED tests/integrations/ds_clients.py::test_data_source_connection[snowflake]
sqlalchemy.exc.ProgrammingError: (snowflake.connector.errors.ProgrammingError)
000666 (57014): Your account is suspended due to lack of payment method.
1 failed, 3 passed, 14 skipped
```

Three things put it outside this change:

- **The diff touches no Snowflake code.** The data-source files here are `aws_cloudwatch_client.py`, `mssql_client.py`, `prometheus_client.py`, `kerberos.py`, `data_source_registry.py` and `data_sources/configs.py`. `snowflake_client.py` is untouched.
- **It reproduces identically on a component PR that touches no data sources at all.** #1060 (document PDF export) fails the same job with a byte-identical error — same test, same `000666`, same `1 failed, 3 passed, 14 skipped`. #1066 fails it too. The common factor is the account, not the diff.
- **The error comes from Snowflake's service, not from our code.** `000666 / 57014` is returned by the account before the query runs; the previous umbrella (#1058) recorded the same failure and reached the same conclusion.

There is no code fix to port. The remedy is on the Snowflake account (a payment method), or gating `test_data_source_connection[snowflake]` on usable credentials the way the other 14 skipped connectors already are — that gating is a change to the integration suite rather than to this release, so I haven't widened this PR with it.

Re-running the failed job once to confirm, per the standing convention. Everything else on this head is green or still running; the rest of the PR is unaffected.

---

yochze:
Correction to the previous comment: I can't re-run the job — both `rerun-failed-jobs` and `rerun` return `403 Resource not accessible by integration`. So don't wait on a re-run from me; it isn't coming.

It doesn't change the conclusion, and the re-run was only ever going to confirm what the cross-PR evidence already shows more strongly: the identical failure on #1060, which touches no data-source code at all. Someone with Actions write access can re-run it if independent confirmation is wanted.

I'll keep watching this PR until it's merged or closed.

---

PR #1078, opened 2026-09-06
Add default-enabled skills to catalog with startup backfill

yochze:
## Summary

This PR introduces the ability to mark skills in the catalog as "default-enabled" so they are automatically installed for every new organization without requiring manual enablement. It includes a startup backfill mechanism to install missing defaults for organizations that predate the feature.

## Key Changes

- **Catalog metadata**: Added two optional frontmatter fields to skill definitions:
  - `order: <int>` — controls position in the catalog listing (lower values first)
  - `default_enabled: true` — marks a skill to be auto-installed for new organizations

- **Default skill installation**:
  - `SkillCatalogService.ensure_defaults_for_org()` — installs all default skills an organization has never had (respects admin disables by checking soft-deleted rows)
  - `SkillCatalogService.backfill_defaults()` — backfills missing defaults for existing organizations on startup, authored by the org's earliest admin
  - `run_default_skill_backfill()` — startup entry point that runs leader-only as a background task

- **Curated catalog order**: Seven skills marked as defaults and given explicit `order` values to appear first:
  - train-agent, audit-instructions, complex-dashboard, migrate-bi-dashboard, usage-review, erd-mermaid, create-evals
  - Remaining skills follow alphabetically with `DEFAULT_ORDER` (1000)

- **Organization creation**: `OrganizationService.create_organization()` now calls `ensure_defaults_for_org()` after admin role assignment to install defaults for new orgs

- **API and UI updates**:
  - Skill catalog endpoint now includes `order` and `default_enabled` fields
  - Frontend displays "Default" label on default-enabled skills with explanatory tooltip
  - Enabled tab sorts skills by catalog rank (defaults first, then hand-authored by recency)

- **Validation**: Added parsing and validation for `order` (must be integer, not bool) and `default_enabled` (must be boolean) in skill file parsing

- **Idempotency**: The backfill is safe to run repeatedly — once a skill has any row (live or soft-deleted), it's skipped, ensuring admin disables persist across restarts

- **Internationalization**: Added "Default" label translations across all supported locales

## Implementation Details

- Soft-deleted rows act as the signal that an admin disabled a default; the feature respects this by never re-enabling a skill that has ever had a row
- Failures during default installation are logged but don't block org creation or startup
- The backfill runs as a background task on the scheduler leader to avoid race conditions in multi-worker deployments
- Default installs are byte-for-byte identical to manually-enabled skills (same instruction row, published status, etc.)

https://claude.ai/code/session_01Gj2Fr6KMVTRmYTjsEdXFUe


yochze:
**CI note: `e2e-tests (sqlite)` failure on the first two heads**

Two distinct causes, both fixed on this branch now:

1. **Nine instruction-count tests** (`test_instruction.py`, `test_instruction_light_view.py`, `test_instruction_activity.py`, `test_instruction_main_build_integrity.py`, `rbac/test_instruction_pending_carryover.py`) assumed a fresh organization has zero instructions. With this PR a new org starts with the seven default skills. Fixed in the second commit: an autouse `no_default_skills` fixture keeps e2e orgs empty, and `@pytest.mark.default_skills` opts the skill-catalog tests back into the real creation path.

2. **`test_seat_cap_autoprovision.py::test_ldap_ensure_memberships_fills_up_to_cap`** is not this PR's: it fails identically on `main` (e2e run for the #1076 merge, 1 failed / 1344 passed, same `AttributeError: 'LDAPGroupSyncService' object has no attribute 'connection'`). The test builds the service with `object.__new__` and the LDAP hardening in #1067 made `_ensure_org_memberships` read `self.connection.provider_id`. No fix PR was open, so the third commit here stubs the connection in that test. It no-ops once `main` carries the same change.

---

PR #1079, opened 2026-09-06
Artifact version management: parent identity table + per-artifact version chains

r0533130357-dev:
# Artifact version management: parent identity table + per-artifact version chains

## Problem

Every row in `artifacts` was one VERSION of a dashboard/deck/doc. Nothing related versions of the same deliverable: `title`/`mode` were copied per row, and `version` was minted independently by 15 different write sites. Observable bugs:

- **"Use this version" on a doc minted v8** because `max(version)` ran over the whole report (a dashboard chain's numbering leaked into the doc's).
- **Two dashboards in one report merged into a single summary row** with interleaved version numbers (ChatSummary grouped by mode).
- A "rebuild from scratch" wrote **v1** even though it replaced an existing dashboard.
- Every "latest" reader did `ORDER BY created_at DESC` with no notion of *which* artifact.

## Design

Two tables (per-deliverable identity split from history):

- **`artifacts`** (new parent): `report_id`, `organization_id`, `created_by`, `mode`, `title` — one row per deliverable. Renaming an artifact renames all its versions; a version can never change kind.
- **`artifact_versions`** (the old table, renamed — **same ids**): gains `artifact_id` FK NOT NULL and `UNIQUE(artifact_id, version)`; loses `title`/`mode` (readable via correlated `column_property`, declared `expire_on_flush=False`).

Version ids never change, so every soft reference stays valid: tool_executions JSON, fork_asset_refs, session events, `thumbnails/{id}.png`, `pdfs/{id}.pdf`, shared links.

**Numbering now has exactly one owner**: `new_artifact()` / `new_version()` in `artifact_service`. All 15 write sites (AI tools, MCP tools, REST routes, forking, seeding, test fixtures) go through them. `new_version` = `max(version)` within the parent + 1 (soft-deleted rows keep their numbers reserved — the UNIQUE spans them).

**Planner contract (D3)**: `create_artifact` defaults to a NEW artifact (v1, its own summary row) even when the report already has one. Only an explicit rebuild ("start over", "redesign completely") passes `replaces_artifact_id` and lands as the next version of that artifact; a mismatched target falls back to a new artifact and says so in the tool observation.

## Migration (`artver01`) — existing customer data

- Runs automatically on upgrade (start.sh → `alembic upgrade head`); transactional on Postgres (all-or-nothing).
- Backfill is deliberately cheap (**no lineage guessing**): every pre-migration version row gets a parent of its own. Wrong merging would be irreversible; non-merging is only cosmetic and reversible — chains can be merged later by re-pointing `artifact_id`.
- `deleted_at` is copied to parents so soft-deleted versions don't resurface as live artifacts.
- Full `downgrade` restores the exact pre-migration schema (verified byte-for-byte on a copy of a real 85-row database, plus an upgrade/downgrade round-trip test through the real alembic chain).
- Index choreography: `ix_artifacts_*` names are schema-global; old ones are dropped up front so the parent table can reclaim them; PG pkey index renamed (SQLite has no ALTER INDEX RENAME).

**Legacy UX is protected**: ChatSummary groups by `artifact_id`, but single-version groups merge by kind+title, so an old 7-edit dashboard still renders as one row with version chips instead of exploding into 7 rows. "Use this version" hides only when the selection is the newest row of its MODE — an artifact_id-only rule would have hidden the revert button from all pre-migration data (every old version is "latest of itself").

## API / UI

- `ArtifactSchema` / `ArtifactListSchema` gain `artifact_id` (additive; `id` stays a version id).
- ChatSummary: one row per artifact (fixes the two-dashboards-merged bug).
- ArtifactFrame: "Use this version" per-mode rule (fixes the button showing on the newest doc).
- Public share page prefers the newest dashboard/deck over a newer doc (same rule as `get_latest_by_report`).
- `artifact_count` now counts artifacts (parents), not versions, consistently on both list and detail paths.

## Notable fixes found during review/testing

- **`expire_on_flush=False` on the title/mode read-throughs**: any post-creation UPDATE-flush of a version row (thumbnail copy, screenshot, pptx path, status) expired the column_properties, and the next access raised MissingGreenlet — this crashed "Use this version" on every artifact that has a thumbnail (i.e. essentially all real data; invisible in tests whose fixtures had no thumbnail file — the REST test now seeds one).
- `edit_artifact` no longer uses `db.get()`: `read_report` leaves a `load_only`-partial instance in the same agent session's identity map, and `get()` returned it as-is → MissingGreenlet on `.content`.
- Deleting the last live version soft-deletes its parent, so identity scans (has_artifacts / artifact_mode filters) stop matching.
- The tools' "Edit (vN)" spec labels and observations report the number actually minted instead of predicting `version + 1` (wrong whenever the source isn't the top of its chain).
- The backfill inserts timestamps through untyped binds (SQLite returns strings from raw SELECT; typed DateTime binds reject them — caught by the new backfill test; the suite's chain replay runs on an empty DB and couldn't see it).
- `alembic/env.py` now registers the artifact models (autogenerate would otherwise propose dropping both tables).

## Testing

- Unit: full suite green on SQLite (4,035 passed) including the complete migration-chain replay; artifact suites green on Postgres.
- e2e (SQLite + Postgres): docs, RBAC shared artifacts, rerun/refresh, HTML/PDF export, MCP, report/project listings, `replaces_artifact_id` tool behavior, REST duplicate/list.
- New: `test_artver01_backfill` (backfill in isolation + full upgrade/downgrade round trip), versioning-semantics tests (independent same-mode chains, soft-delete number reservation, rename scope, constructor guard).
- Migration verified on a real database: 85 rows → 85 parents, 0 nulls/orphans/mismatches; downgrade/upgrade round-trip on a copy.
- Manual QA on `yarn devs`: version revert on legacy + new artifacts, summary grouping, planner new-vs-rebuild, public page, exports, thumbnails.

## Out of scope (documented, pre-existing or deferred)

- Version dropdown doesn't group by artifact (two dashboards can both read "v1").
- The report page's artifact list isn't refreshed after create/revert (pre-existing refresh gap).
- ReadReportTool widget counts version rows as "artifacts".
- `DELETE /artifacts/{id}` has no UI anywhere (pre-existing).
- Per-dashboard sharing, parent deletion UI, URL deep-links to a parent, `current_version_id` pointer, legacy lineage reconstruction.


frontend/components/report/ChatSummary.vue
```diff
@@ -274,23 +274,45 @@ defineExpose({ reloadNotes: loadNotes })
...
+  }
+  // Pass 2: pre-migration history has one parent PER VERSION (the backfill
+  // deliberately refused to guess lineage), so artifact_id alone would
+  // explode an old 7-edit dashboard into 7 rows. Single-version groups merge
+  // by kind+title — how those rows relate historically — while real chains
+  // keep their own identity.
+  const merged = new Map<string, any[]>()
+  for (const versions of byArtifact.values()) {
+    const key = versions.length === 1
+      ? `single:${versions[0]?.mode || 'unknown'}::${versions[0]?.title || ''}`
```
yochze, line 307: **Pass 2 reintroduces the "two dashboards merged into one row" bug for the most common post-migration case.**

Single-version groups are re-keyed by `mode::title`. But under the new contract the planner creates a *new* artifact (v1) by default and tends to reuse the title, and the add-visualization route always scaffolds its dashboard as `"Dashboard"`. So two live dashboards — "Sales Dashboard" v1 (chain A) and "Sales Dashboard" v1 (chain B) — both land on `single:page::Sales Dashboard`, render as one row labelled "v1" with the other dashboard hidden under `older` as a second "v1", and then split back into two rows the moment either gets a v2.

The historical case this pass is meant for (one parent per version from the backfill) is distinguishable from a genuinely new artifact: only merge singles that were created by the backfill (e.g. have the backend flag backfilled parents, or key on a marker the migration sets), rather than on title.

---

backend/app/services/artifact_service.py
```diff
@@ -1,16 +1,141 @@
...
+        content=content,
+        generation_prompt=generation_prompt,
+        completion_id=completion_id,
+        status=status,
+        version=await next_version_number(db, source),
+        **version_fields,
+    )
+    db.add(version)
+    if title is not None and title != source.title:
+        parent = await db.get(Artifact, str(source.artifact_id))
```
yochze, line 129: **`db.get(Artifact, …)` loads the parent with its default loaders, so `Artifact.report` (`lazy='selectin'`) fires and cascades into the Report's whole selectin graph.**

The permissions decorator has already put the Report in the session via `lazyload('*')`, so its relationships are unloaded and the post-loaders for `Artifact.report` → `Report.queries/widgets/completions/…` → `Query.steps` (step data JSON) all run — a multi-query, potentially multi-MB hydration on every rename or delete (`PATCH /artifacts/{id}` with a title, `DELETE /artifacts/{id}`, `edit_artifact`/`edit_doc` with a new title). Same pattern at ~L332 and ~L373.

Suggest `select(Artifact).options(lazyload('*')).where(Artifact.id == …)`, or a bulk `update(Artifact).where(...).values(title=...)` / `.values(deleted_at=now)` statement.

Related: the bare `db.refresh(version)` here and at ~L75 also invokes the version's selectin loaders (report/user/organization/completion) — 4 extra SELECTs per created version, newly added to the fork path which previously only flushed. `db.refresh(version, attribute_names=[...])` bounds it to what the read-throughs need.

---

backend/app/services/report_service.py
```diff
@@ -2732,22 +2767,18 @@ async def get_reports(
                 # Starred state for the current user
                 report_schema.is_starred = str(report.id) in starred_ids
 
-                # Compute unique artifact modes for this report
+                # Compute unique artifact modes for this report (parents)
                 report_schema.artifact_modes = list(set(
                     a.mode for a in (report.artifacts or []) if a.mode
```
yochze, line 2772: **Identity-level scans still count soft-deleted parent rows, so the parent soft-delete added in `ArtifactService.delete` doesn't make "has artifacts" / counts stop matching.**

Delete a report's only artifact version via `DELETE /artifacts/{id}`: the parent gets `deleted_at` set and `GET /reports/{id}` reports `artifact_count=0`, but on the list path:

- `artifact_count = len(report.artifacts)` (~L2754) — the relationship has no `deleted_at` filter → still 1
- `artifact_modes` here still lists `"page"`
- the `has_artifacts=yes` filter (~L2437/L2444, `select(Artifact.report_id)` with no `deleted_at` predicate) still matches
- `project_service.py:781` still counts the report as a dashboard

Add `Artifact.deleted_at.is_(None)` to those subqueries and derive the list-path count/modes from live parents only.

---

backend/app/services/artifact_service.py
```diff
@@ -1,16 +1,141 @@
...
+async def next_version_number(db: AsyncSession, source: ArtifactVersion) -> int:
+    """The number the next version after ``source`` gets.
+
+    max(version) over the parent artifact's rows, plus one. Soft-deleted
+    rows count too: uq_artifact_versions_artifact_version spans them, so a
+    number a deleted row still holds must never be reissued. Never derived
+    from ``source.version + 1``: editing an older version would otherwise
+    mint a number that already exists further up the chain.
+    """
+    max_version = (await db.execute(
```
yochze, line 88: **Read-then-insert with no lock or retry; with the new `UNIQUE(artifact_id, version)` a concurrent second writer now fails with an unhandled `IntegrityError` instead of the previous silent duplicate.**

Scenario: REST add-visualization (or a double-clicked "Use this version") runs while an agent `edit_artifact` on the same artifact is between `max(version)` and flush. Both compute N+1; the second flush raises inside `new_version`, the route returns 500, and the tool's exception handling leaves the session in a failed transaction.

Lock the parent row first (`select(Artifact).where(Artifact.id == …).with_for_update()` — a no-op on SQLite, serialises on PG), or catch `IntegrityError` and retry once.

---

backend/alembic/versions/artver01_normalize_artifact_versions.py
```diff
@@ -0,0 +1,224 @@
...
+            'created_at': r.created_at,
+            'updated_at': r.updated_at,
+            'deleted_at': r.deleted_at,
+        })
+        links.append({'aid': parent_id, 'vid': r.id})
+
+    parents_t = _parents_table_lite()
+    for i in range(0, len(parents), _CHUNK):
+        conn.execute(sa.insert(parents_t), parents[i:i + _CHUNK])
+    update_sql = sa.text("UPDATE artifact_versions SET artifact_id = :aid WHERE id = :vid")
```
yochze, line 111: **The backfill pulls every version row into Python, mints UUIDs client-side, and issues one UPDATE per row via executemany — a two-statement SQL backfill does the same job.**

On a large table this is O(n) UPDATE round-trips inside the migration transaction (psycopg2 executemany is per-statement), plus the `_parents_table_lite` untyped-column workaround that only exists because SQLite returns timestamp strings from the raw SELECT.

Since each version gets its own parent, reuse the version's id as the parent id:

```sql
INSERT INTO artifacts (id, report_id, organization_id, created_by, mode, title, created_at, updated_at, deleted_at)
SELECT id, report_id, organization_id, user_id, mode, title, created_at, updated_at, deleted_at
FROM artifact_versions;

UPDATE artifact_versions SET artifact_id = id;
```

Dialect-neutral, no Python loop, and `_parents_table_lite` / `_CHUNK` go away.

---

yochze (commented):
Review of the artifact/version normalization. The migration's index and PK choreography checks out against the original `n9o0p1q2r3s4` names, there's a single alembic head, and no stale constructors or version-column references survive outside the factory. Nothing crash-level, but five findings inline, ranked most severe first:

1. `ChatSummary.vue` pass-2 merge collapses two distinct live single-version artifacts of the same mode+title (reintroduces the bug the PR fixes).
2. `db.get(Artifact, …)` in new_version/update/delete triggers the Report's full selectin graph on every rename/delete.
3. Soft-deleting the parent doesn't stop list-path counts, `artifact_modes`, `has_artifacts`, or the project dashboard count from matching.
4. `next_version_number` is read-then-insert with no lock; the new unique constraint turns races into unhandled `IntegrityError`s.
5. The migration backfill loops in Python; two SQL statements do the same job.

Checked and deliberately not flagged: `expire_on_flush=False` on the read-through properties, the `replaces_artifact_id` contract, and `read_report`'s `load_only` column set.

---

PR #1087, opened 2026-09-07
Add Kubernetes connector with setup-guide connect form

Dimss:
Read-only Kubernetes data source (category infra, enterprise): 27 fixed tables (workloads, nodes, events, networking incl. ingress classes / network policies / endpoint slices, storage incl. persistent volumes / storage classes / CSI drivers), discovered crd::<group>/<Kind> tables, metrics-server usage tables and pod logs, driven by a JSON query spec. Uniform labels/annotations columns with last-applied stripping and env literal redaction; no Secrets, no exec/proxy paths, GET only.

One auth mechanism: the bearer token of a long-lived service-account token Secret, pasted as an access file printed by
tools/kubernetes/print_access_file.sh (kubeconfig-shaped, strictly parsed — exec plugins / client certs / auth providers rejected by name). RBAC manifest grants cluster-wide read-only.

Connect form: registry entries may carry a numbered `setup_guide` (SetupStep) rendered above the fields with copy buttons, localized via data.setupGuides.<type>; step 1 embeds the RBAC manifest inline as a heredoc so the admin reads exactly what they grant. Copy falls back to execCommand on plain-http deployments (navigator.clipboard is secure-context only) and reports failure. The "Require user authentication" toggle is hidden for types with no user-scoped auth variant (new `supports_user_auth` on the fields response). A stale test result clears on input rather than change, so the first click after correcting a paste is no longer swallowed.

Transport fixes found live: send the full "Bearer <token>" header (the generated client dropped api_key_prefix on kubernetes 36.x, making every call anonymous); request pod logs with "Accept: text/plain, */*" (a bare text/plain gets 406 from some API servers, e.g. microk8s 1.35).

Tests: unit suite over fixtures captured from the kind sandbox; k3s testcontainer integration entry; end-to-end plan + results against a real cluster (docs/feedback-loops/kubernetes-connector-e2e-2026-09-07.md) with reusable scripts under tools/agent/. Sandbox under tools/kubernetes/. Design doc, feedback loop and UI evidence included.

Claude-Session: https://claude.ai/code/session_0122973e2MFnkwiUwX6zXDHs


backend/app/data_sources/clients/kubernetes_client.py
```diff
@@ -0,0 +1,1689 @@
...
+            try:
+                os.remove(path)
+            except OSError:
+                pass
+
+    @staticmethod
+    def _assert_path_allowed(path: str) -> None:
+        if not isinstance(path, str) or not path.startswith("/"):
+            raise ValueError("`path` must be an absolute API path such as /apis/apps/v1/deployments.")
+        segments = [s.lower() for s in path.split("?", 1)[0].split("/") if s]
```
yochze: **Path guard is bypassable with URL encoding.** `_assert_path_allowed` compares literal segments, but the API server routes on the percent-decoded path. The module docstring says this guard, not RBAC, is what keeps Secrets out (the token is cluster-wide read-only), so this matters.

```json
{"path": "/api/v1/namespaces/default/%73ecrets"}
```

Segments become `['api','v1','namespaces','default','%73ecrets']`, nothing matches `_FORBIDDEN_SEGMENTS`, and the request goes through. `ApiClient` concatenates `host + resource_path` verbatim and urllib3 leaves valid percent-escapes intact, while Go's `net/http` hands the apiserver the decoded path. Same for `default%2Fsecrets`, and for the `name` / `namespace` spec fields, which are interpolated into paths before this check runs.

Suggested fix: `urllib.parse.unquote` each segment before comparing, and reject any `%` or `?` in `name`, `namespace`, and `path` outright.

---

backend/app/data_sources/clients/kubernetes_client.py
```diff
@@ -0,0 +1,1689 @@
...
+        kept = []
+        for o in objs:
+            last = parse_rfc3339(o.get("lastTimestamp") or o.get("eventTime") or _meta(o).get("creationTimestamp"))
+            if last is None or last >= cutoff:
+                kept.append(o)
+        return kept
+
+    def _query_path(self, api, spec: dict, limit: int) -> pd.DataFrame:
+        path = spec["path"]
+        params = dict(spec.get("params") or {})
```
yochze: **Forwarding `params` verbatim allows unbounded streaming.** `watch=true` on any list path, or `follow=true` on a pod log path, turns the request into a chunked stream that `resp.data` buffers until the server closes it.

```json
{"path": "/api/v1/events", "params": {"watch": "true", "limit": 1}}
```

`limit` in params skips the pagination loop, so `_get` is called with `watch=true`. `rest.py` builds `urllib3.Timeout(total=request_timeout)`, but urllib3's read/total timeouts only measure time between socket reads, not the total request duration. A cluster emitting events every few seconds streams for the apiserver's watch timeout (30 to 60 minutes by default) while the whole body accumulates in memory. `follow=true` on a chatty container's log path never returns. The worker thread running the query is pinned the entire time.

Suggested fix: strip `watch`, `follow`, and `timeoutSeconds` from params. Also reject `?` in `path`, since `_assert_path_allowed` strips it before checking but `_get` sends it on the wire.

---

frontend/components/datasources/ConnectForm.vue
```diff
@@ -111,7 +136,7 @@
           </template>
 
           <div v-if="showRequireUserAuth && (isCreateMode || isCreateConnectionOnly || isConnectionEdit)" class="flex items-center gap-2 mb-2 mt-4">
-            <UToggle color="blue" v-model="require_user_auth" @change="clearTestResult()" />
+            <UToggle color="blue" v-model="require_user_auth" @input="clearTestResult()" />
```
yochze: **`@input` never fires on `UToggle`.** Nuxt UI 2.16 `Toggle.vue` declares `emits: ["update:modelValue", "change"]` and has no `input` event, so switching this from `@change` silently disables `clearTestResult` for this control.

Repro: run Test connection, see the green result, then flip "Require user authentication". The stale success message stays on screen, which is the exact case the PR description says this change was meant to fix.

The native `<input>` / `<select>` wrappers are fine with `@input`. This one needs to stay `@change` or use `@update:model-value`.

---

backend/app/data_sources/clients/kubernetes_client.py
```diff
@@ -0,0 +1,1689 @@
...
+        crds = self._list(api, "/apis/apiextensions.k8s.io/v1/customresourcedefinitions", limit=MAX_ROWS)
+        specs = self._crd_specs(crds)[:CRD_PROBE_CAP]
+        emitted: List[Table] = []
+        skipped: List[str] = []
+        total = len(specs)
+        for i, spec in enumerate(specs):
+            if progress_callback:
+                progress_callback("custom resources", spec["table"], i, total)
+            try:
+                probe = self._list_objects(api, spec["api"], spec["plural"], spec["namespaced"], None, {}, limit=1)
```
yochze: **CRD population probe ignores the namespace allowlist.** This passes `namespaces=None`, so the probe is cluster-wide while every later query goes through `_resolve_namespaces` and is scoped to the allowlist.

Example: connection configured with `namespaces='payments'`, cert-manager Certificates exist only in `kube-system`. Discovery finds one, emits `crd::cert-manager.io/Certificate` with `spec_keys` taken from an object outside the allowlist, and every query the agent runs against that table returns empty.

Suggested fix: pass `list(self.namespaces) or None` here, the same way `_resolve_namespaces` does.

---

backend/app/data_sources/clients/kubernetes_client.py
```diff
@@ -0,0 +1,1689 @@
...
+        else:
+            pods = self._list_objects(api, "/api/v1", "pods", True, namespaces, params, limit=self.max_log_pods)
+        if not pods:
+            raise ValueError("No pod matched the logs query" + (f" (pod={pod})" if pod else f" (label_selector={selector})") + ".")
+        pods = pods[: self.max_log_pods]
+
+        tail = min(int(spec.get("tail_lines") or self.log_tail_default), self.log_tail_max)
+        since = parse_duration_seconds(spec.get("since"))
+        previous = bool(spec.get("previous"))
+        grep = re.compile(spec["grep"]) if spec.get("grep") else None
```
yochze: **Invalid `grep` leaks a raw `re.error`.** Every other spec problem (bad JSON, unknown table, missing pod) raises a descriptive `ValueError` naming the field. A pattern like `"error("` raises `re.error('missing ), unterminated subpattern')` before any API call, so the agent sees a bare regex message with no hint that `grep` is the culprit.

Suggested fix: wrap the compile in `try/except re.error` and raise a `ValueError` that names the `grep` key.

---

frontend/components/datasources/ConnectForm.vue
```diff
@@ -201,12 +226,62 @@ const emit = defineEmits<{ (e: 'submitted', payload: any): void; (e: 'success',
...
+  if (!ok) {
+    try {
+      const ta = document.createElement('textarea')
+      ta.value = code
+      ta.setAttribute('readonly', '')
+      ta.style.position = 'fixed'
+      ta.style.top = '-1000px'
+      document.body.appendChild(ta)
+      ta.select()
+      ok = document.execCommand('copy')
```
yochze: **Temp textarea leaks if `execCommand` throws.** `removeChild` only runs after `execCommand` returns. In a browser where `document.execCommand('copy')` throws and `navigator.clipboard` is absent (plain-http on-prem, the case the comment above describes), the throw lands in the empty catch and the element stays attached. Each Copy click then appends another off-screen textarea holding the RBAC manifest to the body for the life of the page.

Suggested fix: move `document.body.removeChild(ta)` into a `finally`.

---

backend/app/data_sources/clients/kubernetes_client.py
```diff
@@ -0,0 +1,1689 @@
...
+        # Why a claim is Pending: claim → class → driver
+        pvc = client.execute_query('{"table": "persistent_volume_claims", "namespace": "payments", "field_selector": "status.phase=Pending"}')
+        sc = client.execute_query('{"table": "storage_classes"}')
+        ```
+        """
+
+
+# Alias so dynamic naming ("Kubernetes" → "KubernetesClient") and the explicit
+# client_path both resolve to the same class.
+KubernetesClient = KubernetesClient
```
yochze: **No-op self-assignment.** The comment says this is an alias so dynamic naming resolves, but the class already has this exact name, so the statement changes nothing and reads as an unfinished rename. Remove it, or bind the differently-spelled alias that was intended.

---

yochze (commented):
Review of the Kubernetes connector. Seven inline findings, most severe first:

1. `_assert_path_allowed` compares literal segments, so the Secrets / exec / proxy guard is bypassable with percent-encoding.
2. `_query_path` forwards `params` verbatim, so `watch=true` or `follow=true` opens an unbounded stream the worker buffers in memory.
3. `UToggle` in ConnectForm switched to `@input`, which Nuxt UI 2.x Toggle never emits, so the stale test result no longer clears.
4. CRD population probing ignores the namespace allowlist and advertises tables that always return empty.
5. An invalid `grep` pattern raises a raw `re.error` instead of a field-naming `ValueError`.
6. The clipboard fallback leaks its temp textarea if `execCommand` throws.
7. `KubernetesClient = KubernetesClient` is a no-op self-assignment.

The first two undermine the connector's own stated security boundary and are worth fixing before merge. Verified against the PR head, `kubernetes==36.0.3`, urllib3, and `@nuxt/ui@2.16.0`.

Checked and found fine: `supports_user_auth` toggle visibility for existing types, `setup_guide` availability in production, the hidden tuning fields, locale-key completeness across all ten locales, the new watcher's declaration order, the Bearer-header auth approach, and nanosecond log timestamps.

---

backend/app/data_sources/clients/kubernetes_client.py
yochze: Confirmed against a real `kube-apiserver` v1.31.0 (etcd, RBAC on, the reader token from `tools/kubernetes/rbac.yaml`, a Secret `db-creds` seeded in `default`). With the PR head:

```
{"path": "/api/v1/namespaces/default/%73ecrets"}      -> 1 row, DATA={"password":"U3VwZXJTZWNyZXQxMjM=", ...}
{"path": "/api/v1/namespaces/default%2Fsecrets"}       -> 1 row, DATA={"password":"U3VwZXJTZWNyZXQxMjM=", ...}
```

Fixed in 8f6d15b on `claude/review-pr-1087-7752lk`: path segments and the `name` / `namespace` / `pod` spec values are restricted to `[A-Za-z0-9][A-Za-z0-9._-]*` (no `%`, `/`, `?`, `#`, `..`), and `?`/`#` in `path` are refused with a pointer to `params`. Rejecting the characters is simpler than decoding and there is no legitimate Kubernetes name that needs them. Both encoded forms, `../` in `name`, and `%2F` in `namespace` now raise a `ValueError` before any request; legitimate paths, names and CRD tables still work. Unit tests added for each form.

---

backend/app/data_sources/clients/kubernetes_client.py
yochze: Confirmed against a real apiserver with an event generator creating one Event every 1.5 s: `{"path": "/api/v1/namespaces/payments/events", "params": {"watch": "true", "limit": 1}}` on the PR head was still running after 20 s with `request_timeout=8` (the read timeout never fires while chunks keep arriving).

Fixed in 8f6d15b on `claude/review-pr-1087-7752lk`: `watch`, `follow`, `timeoutSeconds`, `allowWatchBookmarks` and `sendInitialEvents` (case-insensitive) are refused with a `ValueError` explaining they open a stream, and non-object `params` are rejected. Query strings in `path` are refused as part of the path fix. Verified through the backend's own `construct_client` from the stored connection: both `watch=true` and `follow=true` on a log path now return immediately.

---

frontend/components/datasources/ConnectForm.vue
yochze: Correction after testing this in the running app: there is no stale-result bug. `ConnectForm.vue` also has `watch(require_user_auth, () => clearTestResult())` (present on the base branch and on this PR), so the toggle clears the result through the watcher regardless of the inline listener. I verified with Playwright: a failed PostgreSQL test result disappears on flipping the toggle with both `@input` and `@change` in place.

What remains true is that `UToggle` never emits `input`, so the `@input` listener is dead code. 8f6d15b on `claude/review-pr-1087-7752lk` restores `@change` for consistency only. Downgrading this to a nit; no behaviour change either way.

---

backend/app/data_sources/clients/kubernetes_client.py
yochze: Confirmed against a real apiserver with a `cert-manager.io/Certificate` only in `kube-system` and a `Widget` in `payments`. A client with `namespaces="payments"` on the PR head advertised both `crd::cert-manager.io/Certificate` and `crd::example.com/Widget`, and querying the Certificate table returned 0 rows.

Fixed in 8f6d15b on `claude/review-pr-1087-7752lk`: the probe passes `list(self.namespaces) or None`. Verified end to end through the API: a connection created with `namespaces: payments` indexes 26 tables with only `crd::example.com/Widget`, while the unscoped connection indexes 27 with both. Unit test added.

---

backend/app/data_sources/clients/kubernetes_client.py
yochze: Fixed in 8f6d15b on `claude/review-pr-1087-7752lk`: the compile is wrapped and raises `ValueError("`grep` is not a valid regular expression (missing ), unterminated subpattern at position 5): error(")`. Verified against the live cluster and covered by a unit test that also asserts no log request is issued.

---

frontend/components/datasources/ConnectForm.vue
yochze: Fixed in 8f6d15b on `claude/review-pr-1087-7752lk`: `select()` and `execCommand` run inside `try { … } finally { document.body.removeChild(ta) }`. Verified in the real form with Playwright, `navigator.clipboard` removed and `execCommand` stubbed to throw: two Copy clicks leave 0 `textarea[readonly]` elements on `body` and the button shows "Copy failed — select the text".

---

backend/app/data_sources/clients/kubernetes_client.py
yochze: Removed in 8f6d15b on `claude/review-pr-1087-7752lk`. `resolve_client_class("kubernetes")` still resolves to the class by its own name; the registry test asserts it.

---

yochze:
Follow-up on the review: the fixes are in commit 8f6d15b on branch `claude/review-pr-1087-7752lk` (based on this PR's head `c808ded`), ready to cherry-pick or merge into this branch.

**Reproduced against a real control plane** (`kube-apiserver` v1.31.0 + etcd, RBAC on, `tools/kubernetes/rbac.yaml` applied, reader token minted for `bagofwords-reader`), before and after:

| Finding | PR head | With 8f6d15b |
|---|---|---|
| `/api/v1/namespaces/default/%73ecrets` | returns the Secret's `data` | `ValueError`, no request |
| `/api/v1/namespaces/default%2Fsecrets` | returns the Secret's `data` | `ValueError`, no request |
| `name: ../../../../api/v1/namespaces/default/secrets/db-creds` | refused (caught by segment check) | `ValueError` naming `name` |
| `params: {watch: true, limit: 1}` with events flowing | still running after 20 s | `ValueError` immediately |
| CRD probe with `namespaces=payments`, Certificate only in `kube-system` | table advertised, always 0 rows | not advertised; Widget in `payments` still is |
| `grep: "error("` | raw `re.error` | `ValueError` naming `grep` |
| Copy button with no `navigator.clipboard` and a throwing `execCommand` | (not reproduced separately) | 0 leaked textareas after 2 clicks (Playwright) |

**Also verified through the app**: the Kubernetes connection was created through the real connect form (`Connected successfully. Found 27 tables.`), the stored connection's encrypted credentials rebuild the client via `connection_service.construct_client`, and the same query battery through that client gives the results above. The only leg not exercised is an LLM-driven chat turn, because the sandbox's Anthropic key has no credits.

**One correction**: the `@input` vs `@change` finding on the require-user-auth toggle is a nit, not a bug. The form's `watch(require_user_auth, …)` already clears the test result, so the stale-result scenario never happens. Thread resolved with that note.

Tests: `tests/unit/test_kubernetes_client.py` passes (138 tests, 25 new cases); ruff findings on the two files are unchanged from this PR's baseline.

---

backend/app/data_sources/clients/kubernetes_client.py
Dimss: Fixed in 881287af. Rather than decoding, every path segment and every `name`/`namespace`/`pod` spec value must now be a plain Kubernetes name (`[A-Za-z0-9][A-Za-z0-9._-]*`); `%`, `/` inside a segment, `?`, `#` and `..` are refused with a ValueError before any request. Both encoded forms from your repro plus `../secrets` in `name` and `a%2Fb` in `namespace` are covered by unit tests, and legitimate paths, names and `crd::` tables were re-verified against a live cluster.

backend/app/data_sources/clients/kubernetes_client.py
Dimss: Fixed in 881287af. `watch`, `follow`, `timeoutSeconds`, `allowWatchBookmarks` and `sendInitialEvents` (case-insensitive) are refused with a ValueError explaining they open a stream, non-object `params` are rejected, and `?` in `path` is refused by the path guard so nothing can be smuggled around `params`. Unit tests assert no request is issued in any of these cases.

frontend/components/datasources/ConnectForm.vue
Dimss: Fixed in 881287af — restored `@change` on the UToggle. Agreed it is a nit given the `watch(require_user_auth)` already clears the result; the `@input` listener was dead code and is gone.

backend/app/data_sources/clients/kubernetes_client.py
Dimss: Fixed in 881287af. The population probe now passes `list(self.namespaces) or None`, the same scope as every later query. A unit test seeds a kind populated only in `kube-system` and asserts a client scoped to `payments` neither probes it cluster-wide nor advertises its table, while an unscoped client still does.

backend/app/data_sources/clients/kubernetes_client.py
Dimss: Fixed in 881287af. The compile is wrapped and raises `ValueError("`grep` is not a valid regular expression (…): 'error('")`; a unit test also asserts no log request goes out for an invalid pattern.

frontend/components/datasources/ConnectForm.vue
Dimss: Fixed in 881287af. `select()` and `execCommand` now run inside `try { … } finally { document.body.removeChild(ta) }`. Verified with Playwright on a plain-http origin (no `navigator.clipboard`) with `execCommand` stubbed to throw: two Copy clicks leave 0 readonly textareas on `body` and the button reports "Copy failed — select the text".

backend/app/data_sources/clients/kubernetes_client.py
Dimss: Removed in 881287af. `resolve_client_class('kubernetes')` resolves the class by its own name, which the registry unit test asserts.

yochze:
Re-verified `881287a` against the same local control plane (kube-apiserver v1.31.0, RBAC, seeded Secret / CRDs / pod) and the unit suite:

- Every attack spec from the review is refused before a request: `%73ecrets`, `default%2Fsecrets`, `?` in `path`, `../` in `name`, `%2F` in `namespace`, `watch=true`, `follow=true`, `grep: "error("`. Legitimate pods, the configmap escape hatch and the Widget CRD table still return rows.
- CRD probe with `namespaces=payments` advertises only `crd::example.com/Widget`; the Certificate in `kube-system` is gone.
- `tests/unit/test_kubernetes_client.py`: 132 passed. Ruff on the two files is unchanged from the PR's baseline.
- ConnectForm changes match what I had verified with Playwright (toggle back to `@change`, `finally` around the textarea).

This supersedes my branch `claude/review-pr-1087-7752lk`; nothing there is needed anymore.

Two small nits, neither blocking:

1. `grep` given as a non-string (`{"grep": 5}`) reaches `re.compile` and raises a raw `TypeError: first argument must be string or compiled pattern`, the same class of leak the `re.error` fix closed. Wrapping in `str(...)` or catching `TypeError` alongside `re.error` covers it.
2. `_check_segment` rejects a non-string `name` such as `{"name": 123}`, which a model may emit for an all-digit pod name (a valid DNS label). Coercing with `str(value)` before matching keeps the same guarantees.

One behaviour worth knowing: the namespace allowlist in the connection config is now validated at construction, so an admin who types `payments/x` or `*` gets a `ValueError` naming `namespace`. Both the stateless and stored test-connection paths catch it and show the message, so the UI degrades cleanly. Previously `*` silently produced an allowlist of `['*']`, so this is an improvement.

---

PR #1089, opened 2026-09-07
Add sslmode support to Postgres connector

ddeanto:
The Postgres data source connector has no SSL/TLS config, unlike most other connectors (e.g. SAP HANA's `encrypt`/`verify_ssl`). This blocks connecting to providers like AWS RDS that reject unencrypted IAM-authenticated logins.

Adds an `sslmode` field to `PostgreSQLConfig`, passed through as `connect_args` to the existing engine pool (same mechanism the app's own internal DB connection already uses). Defaults to `prefer` (psycopg2's own default), so existing connections are unaffected.

Tested locally against a Postgres instance configured to reject unencrypted connections (`hostssl`-only in `pg_hba.conf`) — confirmed `sslmode=require` connects successfully where a plaintext attempt is rejected, and confirmed the default `prefer` still connects fine to a normal instance.


yochze:
Thanks for contributing @ddeanto!

PR #1095, opened 2026-09-08
security(code-exec): run generated code in a sandboxed child process, not the API worker

yochze:
**What:** model-generated Python (create_data / inspect_data and the python-pptx slides path) now runs in a forked, rlimited, Landlock-confined child process with a scrubbed environment; the API worker only brokers queries and web fetches to it over a pipe.
**Why:** it ran through `exec()` inside the uvicorn worker, guarded only by the AST denylist, so one escape reached the encryption key, the app database, every connector's credentials and any Kerberos keytab, and a runaway loop could not be killed.

## Executive summary

For an on-prem customer, the analytics agent writes and runs Python against their data. Until now that code shared a process with everything the deployment holds: `BOW_ENCRYPTION_KEY`, the application database, live connections to every data source, and the SQL Server keytab where SSO is configured. Python denylists are a well-known bypass genre, so a prompt-injected or simply mistaken piece of generated code was one step from decrypting every stored credential, and an infinite loop or memory blow-up took an API worker down with it. After this change, generated code runs in a throwaway process that has no secrets, no application database, a memory cap and a hard time limit, and on current kernels no filesystem beyond its own inputs and no TCP. Nothing changes for users: same tools, same prompts, same results.

## Visuals

**4a. What a customer is exposed to — before vs after**

```mermaid
flowchart LR
  subgraph before[Before: one process]
    U1[User asks a question] --> G1[Generated Python via exec in API worker]
    G1 -->|same heap, same env| S1[(Encryption key, app DB, all connector credentials, keytab)]
    G1 -->|no cap, no kill| W1[API worker hangs or OOMs]
  end
  subgraph after[After: sandboxed child]
    U2[User asks a question] --> G2[Generated Python in a forked sandbox child]
    G2 -->|only execute_query over a pipe| B2[API worker brokers queries]
    G2 -.->|denied: no secrets, no env, Landlock, rlimits, SIGKILL| S2[(Secrets stay in the API worker)]
  end
  style S1 fill:#f8d7da,stroke:#c00
  style W1 fill:#f8d7da,stroke:#c00
  style S2 fill:#d4edda,stroke:#080
  style B2 fill:#d4edda,stroke:#080
```

**4b. Mechanism**

```mermaid
sequenceDiagram
  participant API as API worker (trusted)
  participant Z as fork server (spawned once, scrubbed env, libs imported, no secrets)
  participant C as sandbox child (untrusted)
  API->>API: validate_python_code (AST), wrap clients
  API->>Z: fork request + 3 pipe fds (Unix socket, SCM_RIGHTS)
  Z->>C: fork(); setsid()
  API->>C: job (pickle: code, file attrs, loadables, limits, fs policy)
  C->>C: rlimits, no_new_privs, Landlock (fail-open)
  C-->>API: ready (what was applied)
  C->>C: exec(code); generate_df()
  C->>API: rpc execute_query(client_key, args, kwargs)
  API->>API: QueryCapturingClientWrapper: capture, timeout, quota, rate limit, SQL render
  API-->>C: DataFrame (pickle, trusted direction)
  C-->>API: result as Arrow IPC + stdout (never pickle)
  API->>C: SIGKILL on timeout/cancel (pid reserved until reaped)
  API->>Z: reap pid
```

## Reviewer decision box

| | |
|---|---|
| **Risk** | Medium. Touches the execution path of every generated-code tool; behavior is preserved by tests, but the transport and the fork server are new. |
| **Blast radius** | `StreamingCodeExecutor.execute_code` (create_data, inspect_data, write_csv, MCP variants, entity/saved-query refresh) and `PptxCodeExecutor.execute_pptx_code`. |
| **Behavior change (intended)** | Generated code sees a client proxy that exposes only `execute_query` / `query`; other client attributes raise `AttributeError` with guidance. Error text for user-code exceptions carries the type name (`KeyError: 'x'`). Memory (4 GB) and wall-clock (600 s) caps apply where none existed. The AST rule from 9f910b5 that rejects underscore attributes now exempts `__name__` (see *Deep detail*). |
| **Verified** | CI green on `43938ad` (pre-merge head). Live stack: saved query against the Chinook demo DB via API and UI; child pid distinct from worker; encrypted DB row decrypts to the same rows; negative cases rejected. Fork server: distinct pid per run, ~20 ms per execution after the first, timeout kill and reap, sklearn (BLAS/OpenMP) inside a forked child, spawn fallback. |
| **NOT verified** | The LLM chat path itself (authoring env had no API credit; the entity path exercises the same executor). Landlock syscalls only ran on CI's Ubuntu runners, not in the authoring container (LSM compiled out there). No Windows support (Linux-only mechanisms; the in-process mode remains). CI on the current head (`d29789c`) is in progress at the time of writing. |
| **Where to look first** | `backend/app/ai/code_execution/sandbox/runner.py` (`run_job`, `_Zygote`, `_fs_policy`, `_child_environment`), `sandbox/zygote.py`, `sandbox/child.py` (`_RemoteClient`, `run`), `sandbox/protocol.py` (trust direction of the wire format). |

## Evidence

| Check | Result |
|---|---|
| CI on `43938ad` (unit, e2e sqlite + postgres, playwright, integrations) | all success |
| CI unit-tests on `99fdd1a` (merged main, warm-pool design) | cancelled at the job's 60-min cap — main's own unit job already takes 51–54 min; the per-run interpreter spawn pushed it over. Fixed by the fork server in `6164f27`. |
| New tests `tests/unit/test_code_sandbox.py` | 21 passed (env scrub, no parent memory, client lockdown, timeout kill, memory cap, cancel kill, original-exception propagation, Arrow round trip + fallback, pptx, fork server pid/latency, spawn fallback, validator `__name__` rule, Landlock ABI layout) |
| Existing execution suites run locally against the sandbox | green (129 + 147 on the first head; 76 across the sklearn/sandbox/concurrency modules on the current head) |
| Failures attributable to this PR | 1, fixed in `43938ad`: the first push dropped two re-exports (`READ_TEXT_MAX_CHARS`, `_build_read_text`) a test module imports, breaking collection. |
| Pre-existing failure on main, fixed here | 4 `test_sklearn_sandbox` tests fail on `origin/main` (verified in a clean worktree): 9f910b5 rejects `type(model).__name__`. `d29789c` exempts `__name__`; every other underscore attribute stays forbidden (regression test added). |
| Automated review | no bot threads on this PR at time of writing |

## Context map

- **Key files**: `backend/app/ai/code_execution/sandbox/{runner,zygote,child,protocol,namespace,landlock,config}.py`; wiring in `backend/app/ai/code_execution/code_execution.py` (`_execute_sandboxed`, `_execute_inprocess`) and `backend/app/ai/code_execution/pptx_executor.py`.
- **Convention established**: anything generated code must reach on the trusted side is an RPC method handled in `runner.run_job`; nothing crosses child → parent except JSON headers and Arrow bytes. New injectables go through `sandbox/namespace.py`, which must stay importable without the app (no DB, no settings). The fork server must never receive job data; it only imports and forks.
- **Operator knobs**: `BOW_CODE_SANDBOX`, `BOW_SANDBOX_TIMEOUT_SECONDS`, `BOW_SANDBOX_MEMORY_MB`, `BOW_SANDBOX_CPU_SECONDS`, `BOW_SANDBOX_REQUIRE_LANDLOCK`, `BOW_SANDBOX_ZYGOTE` — documented in `docs/design/code-execution-sandbox.md`.
- **Docs**: `docs/design/code-execution-sandbox.md` (design, guarantees, non-guarantees), `docs/feedback-loops/code-execution-sandbox.md` (reproduce → verify loop).
- **Commits**: `b8871fa` sandbox + wiring + tests + docs; `43938ad` restore re-exports; `b2f9c23` merge main; `99fdd1a` test adapted to main's attribute rule; `6164f27` fork server; `d29789c` `__name__` exemption.
- **Follow-ups**: (1) run the child as a separate container with a no-egress network policy and an optional `runtimeClassName` (gVisor / Kata) — the pipe protocol is the same; (2) network isolation on kernels older than 6.7 depends on that step; (3) changelog entry at release.

## Deep detail

**Root cause.** `execute_code` built a namespace containing the wrapped clients and called `exec` on a thread of the API process. The AST validator shapes model output well but is not a boundary (`getattr`-free attribute walks, `pd.read_*` with a variable path, and similar are documented bypass classes).

**Approach.** Never fork from the API worker. Each API process spawns one fork server with a scrubbed environment that imports numpy/pandas/pyarrow and then only forks; a child is a fork of that clean process and applies its own rlimits, `no_new_privs` and Landlock. Keep every policy-bearing object (client wrappers, `SafeHttpClient`) in the parent and expose them as RPC. Return DataFrames as Arrow IPC with a per-column string fallback for untypeable columns (reported in the log). Re-raise trusted-side exceptions as the original objects when they propagate unchanged, so `QueryTimeoutError`, `UnsafeSQLError` and usage-limit errors keep their handling. Finished children stay zombies until the runner reaps them via the zygote, so the runner's SIGKILL can never hit a recycled pid. Landlock via ctypes (ABI 1–6, masked to what the kernel offers); fail-open by default because many VM kernels lack the LSM, with an env switch to fail closed.

**Roads not taken.** Firecracker/microVM (needs KVM; not available on most on-prem VMware/Hyper-V guests). nsjail/bubblewrap inside the app container (needs `CAP_SYS_ADMIN` or a relaxed seccomp profile on the customer host). Pyodide/WASM (no supported server runtime, weaker pandas story, recent sandbox escape). A warm pool of pre-spawned interpreters (the first design here) hid the 0.7 s import from the request path but not from CPU, and pushed the unit-tests job past its cap.

**Self-introduced issues.** A lint pass before the first push removed two names an existing test imports (fixed in `43938ad`). The warm-pool design made the CI unit job exceed 60 minutes on the merged base (replaced by the fork server in `6164f27`).

**Fork safety.** The zygote holds no Python threads. numpy's OpenBLAS registers `pthread_atfork` handlers, and a forked child running scikit-learn (OpenMP + BLAS) was exercised explicitly. `BOW_SANDBOX_ZYGOTE=0` falls back to one full interpreter per run.

**Cost.** Cold interpreter ≈ 0.7 s; forked child ≈ 5–20 ms; one zygote ≈ 100 MB RSS per API process.

**Not in scope.** Kernel isolation; network isolation on kernels < 6.7; the artifact (JSX) render path, which runs in the browser sandbox and was not touched.

https://claude.ai/code/session_01EnKJfyMSAQn4NxeHPWvm3Q


yochze:
**CI note: two `e2e-tests` failures on this branch are present on `main` itself, not caused by this PR.**

On head `99fdd1a` the `e2e-tests (sqlite)` job finished `2 failed, 1426 passed`:

- `tests/e2e/rbac/test_rbac_training_mode.py::test_agent_creator_can_train_agent_they_created` — expects 403 when a creator switches a report to training after its inaccessible agent was dropped; the API now returns 200.
- `tests/e2e/test_diagnosis_explorer.py::test_summary_line_is_computed_from_the_query` — the explorer summary now carries `unindexed: 0`, which the exact-dict assertion does not expect.

Both reproduce identically on a clean checkout of `origin/main` (`d84aff0`), run with `pytest -m e2e --db=sqlite` on just those two tests, and this PR changes nothing outside `backend/app/ai/code_execution/` and its tests. They appear to come from the combination landed by #1101 (training-mode gating and the diagnosis explorer merged together); no fix is on `main` yet.

Proposed patches, not applied here to keep this PR to the sandbox:

- diagnosis summary: add `"unindexed": 0` to the expected dict in `test_summary_line_is_computed_from_the_query` (the field is legitimate; the assertion is stale).
- training-mode RBAC: decide whether `PUT /reports/{id}` to `training` must 403 when the report retains no trainable agent (test's expectation) or 200 (current behavior after #1102); either the gate in `update_report`'s training check or the test needs to follow.

This PR's other failure, four `test_sklearn_sandbox` tests red on `main` since 9f910b5, is fixed here in `d29789c` because it lives in the validator this PR touches.

---

yochze:
**CI note (unit-tests on `d29789c`): 5 failures, all present on `main`, none in this PR's files.**

The job otherwise passed (`5 failed, 4760 passed, 57:06` — back under its 60-minute cap after the fork server in `6164f27`). The five:

- `tests/unit/test_read_query_observation_code.py` — three tests: `AttributeError: '_FakeDB' object has no attribute 'scalar'` at `app/ai/tools/implementations/read_query.py:266`, the new BOW-source access gate added by 9f910b5, which the test's fake session predates.
- `tests/unit/test_artifact_relationship_loading.py::test_read_query_loads_only_requested_query_graph[query|visualization]` — the same gate issues an extra `Query.report_id` lookup, which the "loads only the requested graph" assertion counts.

All five fail identically on a clean checkout of `origin/main` (`d84aff0`); this PR does not touch `read_query.py` or those tests. Proposed patch (not applied here): give `_FakeDB` a `scalar()` returning the report id, and either whitelist the `Query.report_id` lookup in the relationship-loading assertion or resolve it through the already-loaded query object in `read_query.py`.

Together with the two e2e failures noted above, that is the full set of red on this head that originates on `main`.

---

yochze:
Reviewed head `d29789cd5782fc89380b70b4cd3d8eb13359ed61`. I recommend changes before merging:

1. **[P1] Landlock policy fails on ordinary Linux files** — [`runner.py:153`](https://github.com/bagofwords1/bagofwords/blob/d29789cd5782fc89380b70b4cd3d8eb13359ed61/backend/app/ai/code_execution/sandbox/runner.py#L153)

   `/etc/ld.so.cache` and `/etc/localtime` are in `read_paths`, which grants `FS_READ_DIR` alongside `FS_READ_FILE`. Linux rejects directory rights on files with `EINVAL`; `restrict_self()` then abandons the entire policy. Default deployments consequently lose filesystem and TCP confinement, while strict deployments reject execution. Move these entries to `read_files`. This follows directly from the [kernel validation code](https://github.com/torvalds/linux/blob/v6.8/security/landlock/fs.c#L154-L157).

2. **[P1] Uploaded files become unreadable** — [`namespace.py:65–66`](https://github.com/bagofwords1/bagofwords/blob/d29789cd5782fc89380b70b4cd3d8eb13359ed61/backend/app/ai/code_execution/sandbox/namespace.py#L65-L66)

   `FileService` stores relative paths such as `uploads/files/example.csv`. These pass unchanged into children running from a scratch directory. A reproduction created that CSV relative to the parent working directory and executed `pd.read_csv(excel_files[0].path)`; both fork-server and spawn modes raised `FileNotFoundError`. Resolve file paths in the parent before serialization, consistently with the filesystem allowlist.

3. **[P1] Blocking operations bypass timeout and cancellation enforcement** — [`runner.py:427–431`](https://github.com/bagofwords1/bagofwords/blob/d29789cd5782fc89380b70b4cd3d8eb13359ed61/backend/app/ai/code_execution/sandbox/runner.py#L427-L431)

   After `select()` reports any bytes, `read_message()` blocks until the entire frame arrives without checking the deadline or cancellation event. Synchronous RPC handlers likewise suspend enforcement. A bounded partial-frame reproduction remained blocked after four times its deadline; a three-second stub RPC exceeded a one-second sandbox limit and only raised after 3.03 seconds. CPU limits cannot terminate a child sleeping or waiting on a pipe. Use deadline-aware I/O and independent timeout/cancellation supervision around blocking operations.

4. **[P2] Forked jobs retain the shared fork-server scratch directory** — [`child.py:304–308`](https://github.com/bagofwords1/bagofwords/blob/d29789cd5782fc89380b70b4cd3d8eb13359ed61/backend/app/ai/code_execution/sandbox/child.py#L304-L308)

   The child receives its private scratch directory but never updates its working directory or temporary/cache environment variables inherited from the fork server. With Landlock unavailable, a first job wrote a synthetic CSV using a relative path and a second job successfully read it; both used the fork server's directory, which per-job cleanup does not remove. Once confinement applies, those inherited locations fall outside the writable policy. Set the child's working directory and temporary/cache environment to its private directory before applying restrictions.

**Validation:** Findings 2–4 were reproduced against the PR's sandbox modules in disposable, network-disabled Linux containers using synthetic data and bounded probes. The local Linux kernel reported Landlock ABI 0, so finding 1 was verified against the PR and Linux kernel source, not through an enabled Landlock runtime. This was a focused review, not a full application test-suite run.

yochze:
All four findings confirmed and addressed in `69b995f`:

1. **Landlock policy aborted by file entries** — `_add_path_rule` now masks rights by path type (a plain file only ever gets `READ_FILE`/`WRITE_FILE`/`TRUNCATE`/`EXECUTE`/`IOCTL_DEV`), a refused rule is reported in `rule_errors` and skipped instead of abandoning the policy, and `/etc/ld.so.cache` and `/etc/localtime` moved to `read_files` (with `/etc/resolv.conf`). Since the authoring kernel has no Landlock, `test_landlock_policy_applies_and_confines` now runs the real policy on CI's Ubuntu runners: policy applied with no rule errors, the uploaded file readable, a sibling file and `/etc/hostname` denied, `import sklearn` still working post-confinement, and TCP connect refused on ABI ≥ 4.

2. **Relative upload paths** — `file_to_attrs` resolves `path` with `os.path.abspath` on the parent before it crosses the boundary (same base the in-process code used). Test: a `uploads/files/sample.csv` relative to the parent cwd, read via `pd.read_csv(excel_files[0].path)` in the child.

3. **Blocking I/O and RPC bypassing the deadline** — the result pipe is non-blocking and every frame read goes through a supervised `select` loop (`_read_exact_with_deadline`) that checks the wall clock and the cancel event between chunks; RPC handlers run on a worker thread under the same supervision (`_call_supervised`), so a slow query no longer suspends enforcement — the child is killed at the limit and the handler is abandoned (the query wrappers already bound and cancel their own calls). Tests: a 4 s stub query under a 1 s limit raises `SandboxTimeoutError` in < 2.5 s; a 2-of-8-byte partial header stalls and is killed at the deadline.

4. **Forked jobs sharing the fork server's scratch dir** — `child.run` now `chdir`s into the job's private scratch dir and points `TMPDIR`/`TMP`/`TEMP`/`HOME`/`MPLCONFIGDIR` at it (and resets `tempfile.tempdir`) before applying rlimits/Landlock. Test: two consecutive jobs see different cwds equal to their `tempfile.gettempdir()`, a marker written by the first is not visible to the second, and both dirs are gone afterwards.

Local: 121 sandbox/executor tests green (Landlock test skipped here, ABI 0) plus the 114 query-wrapper tests. CI on `69b995f` is running; the Landlock test is the one to watch since this is its first run on a kernel that has the LSM.

---

yochze:
**CI on `69b995f`** — the Landlock policy ran for real on the Ubuntu runner and held: `test_landlock_policy_applies_and_confines` passed (policy applied with no rule errors, upload readable, sibling file and `/etc/hostname` denied, TCP connect refused, `import sklearn` fine after confinement). pytest finished `5 failed, 4766 passed in 59:11` — the same five `read_query` failures that are on `main` — but the job was cancelled at 60:00 during teardown by its cap. e2e sqlite/postgres and the other jobs are unaffected.

**`f649508`**:
- The supervised RPC loop only noticed a finished handler at its next 250 ms tick; the worker now signals through a wake pipe the supervisor selects on, so fast queries return immediately (deadline/cancel still enforced).
- `unit-tests` `timeout-minutes` 60 → 75, with a comment. `main` alone takes ~52 min; the sandbox adds a fork, Landlock and an Arrow round trip to every generated-code test (~5–7 min), which left the job finishing inside its last minute. Flagging this explicitly since it is a CI config change outside the sandbox itself.

---

PR #1096, opened 2026-09-08
Enforce table activation for delegated (user_required) sources

yochze:
## Summary

Fix a security and correctness issue where agents on delegated data sources (those requiring user authentication) were serving users every table their own credentials could reach upstream, ignoring the agent manager's table activation selection. This affected ~50 connectors with `auth_policy=user_required` including Power BI, Analysis Services, and PostgreSQL with row-level security.

## Root Cause

When a user signs in with their own credentials to a delegated data source, the system creates an overlay of tables they can access. Previously, three code paths ignored the canonical `is_active` flag (the manager's activation selection) and treated every accessible overlay table as active:

1. **Schema context builder** — used by the planner and `create_data` resolution
2. **Query-time metadata map** — Power BI/Analysis Services table→dataset resolution
3. **DAX query validation** — no enforcement that query bodies only reference activated tables

A user-discovered delegated dataset (e.g., a Power BI RLS model) had no canonical row in the service principal's catalog, so the workaround was to ignore the flag entirely. This is now fixed by creating the canonical row on overlay sync.

## Key Changes

- **`schema_context_builder.py`** — The overlay branch now gates each overlay table on its canonical row's `is_active` flag when `active_only=True` (default). Missing canonical rows count as not activated. Relationship targets are recomputed from visible tables so foreign keys never point at hidden tables. With `active_only=False` (management surfaces), inactive rows are emitted flagged as inactive.

- **`data_source_service.py`** — `_attach_stored_table_metadata` now only includes activated canonical rows in the query-target resolution map. The user's overlay can enrich an activated row's metadata but never makes a non-activated row resolvable. Non-activated rows are passed to `attach_blocked_table_metadata` for clients that need to enforce DAX body restrictions.

- **`powerbi_client.py`** — Added `attach_blocked_table_metadata` to track non-activated tables per dataset. `execute_query` now refuses:
  - A non-activated table as a query target (with helpful error listing activated tables)
  - A DAX body that references a non-activated sibling table of the target dataset via lexical analysis (`'Quoted Name'`, bare identifiers in table position)

- **Tests** — Comprehensive unit tests covering:
  - Inactive canonical hides accessible overlay table
  - Activating canonical row surfaces overlay table with user columns
  - Overlay row without canonical row is not activated
  - Manager activates 3, user reaches 2 → agent shows 2
  - Creator activates 4, user2 reaches 10 → user2's agent shows 4
  - User with no upstream access sees nothing
  - Relationship targets restricted to activated tables
  - `active_only=False` emits inactive tables flagged
  - Power BI DAX body enforcement (target and sibling references)

## Implementation Details

- **Activation rule**: A table is in the agent for a user iff `overlay.is_accessible == True` AND `canonical.is_active == True`
- **Canonical row creation**: `_upsert_user_overlay` now creates missing canonical rows on sync, enabling the tables wizard to activate delegated datasets like any other table
- **DAX validation**: Best-effort lexical check ignoring double-quoted string literals; references detected as `'Quoted Name'` or bare identifiers in table position
- **Backward compatibility**: Clients without `attach_blocked_table_metadata` (e.g., Analysis Services) continue to work; the activation gate still applies via the metadata map

Validated live against PostgreSQL 16 with row-level security and a real Entra/Power BI tenant with service principal + delegated users.

https://claude.ai/code/session_018hbM8QuMQn3KQJRYpFyomf


backend/app/data_sources/clients/powerbi_client.py
```diff
@@ -153,6 +155,81 @@ def attach_table_metadata(self, tables: List[Dict]) -> None:
...
+            if table_name.lower() in activated:
+                continue  # same internal name activated via another row
+            esc = re.escape(table_name)
+            quoted = re.compile(r"'" + esc + r"'", re.IGNORECASE)
+            if quoted.search(body):
+                hits.append(schema_name)
+                continue
+            if re.fullmatch(r"\w+", table_name):
+                bare = re.compile(
+                    r"(?<![\w\[\]'\.])" + esc + r"(?=\s*\[|\s*[,)]|\s*$|\s*\n)",
```
yochze: **[P1] Handle bare table references followed by ORDER BY**

With `SalesModel/Customers` activated and sibling `SalesModel/Orders` blocked, `execute_query("EVALUATE Orders ORDER BY [Id]", "SalesModel/Customers")` passes this guard and reaches `_execute_dax_internal`. The lookahead does not recognize a bare table name followed by `ORDER BY`, although this is supported DAX syntax (https://learn.microsoft.com/en-us/dax/orderby-statement-dax). I reproduced this with the execution boundary mocked: `EVALUATE Orders` is blocked, while the sorted form reaches execution. This leaves the inactive-table restriction bypassable whenever the user's upstream credentials can read Orders. Please handle table references through proper token recognition, including query clauses and comments, and add regression coverage for this form.

backend/app/ai/context/builders/schema_context_builder.py
```diff
@@ -391,6 +391,30 @@ async def build(
...
+                # agent beyond what its manager activated. A user-discovered
+                # delegated table gets its canonical row created on sync (see
+                # DataSourceService._upsert_user_overlay), so there is always
+                # a row to activate in the tables wizard; a row that is still
+                # missing is treated as not activated.
+                if active_only:
+                    overlay_tables = [
+                        ot for ot in overlay_tables
+                        if bool(getattr(
+                            canonical_all_by_name.get(getattr(ot, 'table_name', '') or ''),
```
yochze: **[P2] Resolve activation through the overlay's canonical ID**

A Power BI dataset rename can leave an accessible overlay named `SalesV2/Orders` linked by `data_source_table_id` to an activated canonical row named `Sales/Orders`. The existing overlay sync deliberately matches `(datasetId, tableName)` and preserves the canonical display name for service-principal-indexed rows. This new name-only lookup treats that correctly linked table as missing and hides it. I reproduced the state through `get_user_data_source_schema(..., prefetched_tables=...)`: the canonical row remained active and the overlay pointed to its ID, but both the planner schema and `read_user_data_source_schema(active_only=True)` returned empty. Please resolve activation using the canonical ID, with an appropriate fallback for unlinked legacy overlays, and apply the same correction to the new name-based filter in `read_user_data_source_schema`.

yochze (commented):
Found two issues at b9f460c962ea5df5852f9fa35f544ad6adc43edd, detailed inline.

Validation: all 83 tests across test_powerbi_client.py, test_overlay_inactive_canonical_context.py, and test_attach_table_metadata_delegated.py passed. Additional focused reproductions confirmed the DAX guard bypass (with the execution boundary mocked) and the renamed-dataset schema regression (using an in-memory SQLite database and the real overlay sync). Neither case is covered by those tests. No live Power BI queries were run.

backend/app/data_sources/clients/powerbi_client.py
yochze: Fixed in 1170600. The lookahead regex is replaced by a small DAX lexer (`_dax_table_references`): comments (`//`, `--`, `/* */`), string literals, quoted identifiers, bracketed column/measure names and bare words are tokenized, and a table reference is a quoted identifier or a bare word not followed by `(`. That covers every clause position (`EVALUATE T ORDER BY ...`, `START AT`, `DEFINE VAR/MEASURE`, `FUNC(T, ...)`, `T[col]`) while comment and string contents never surface.

Regression coverage added in `TestActivationEnforcement` (ORDER BY / START AT / DEFINE / comment forms, plus negative cases such as `ORDER BY [Orders]`, `-- see Orders`, quoted names inside strings) and `TestDaxTableReferences` for the lexer. Re-verified against the live tenant: `EVALUATE Customers ORDER BY [CustomerID]` via `SalesPush/Sales` is now refused; the activated form with ORDER BY and a query mentioning the sibling only in a comment still run.

---

backend/app/ai/context/builders/schema_context_builder.py
yochze: Fixed in 1170600. Activation now resolves through the overlay's `data_source_table_id` link first, with the name as fallback only for unlinked legacy overlays, in all three places:

- `SchemaContextBuilder` overlay branch: canonical rows are loaded by id as well as by name (`_canonical_for`), and the relationship-target set carries both the overlay and canonical names so FKs still prune correctly across a rename.
- `read_user_data_source_schema(active_only=True)`: filters by active canonical id when linked, by name otherwise.
- `_attach_stored_table_metadata`: an overlay name linked by id to an activated row is added to the query-target map under its own (renamed) name, since that is the name the user's schema context shows; linked to a non-activated row it lands in the blocked map instead.

Tests: `test_renamed_dataset_overlay_linked_by_id_follows_canonical_activation`, `test_id_link_wins_over_a_coincidental_name_match`, and the three id-link cases in `test_attach_table_metadata_delegated.py`.

---

yochze:
CI status on 1170600: `unit-tests` shows cancelled, not failed. The job hit its `timeout-minutes: 60` budget at 98% of the suite with zero failures in the log (`##[error]The operation was canceled.` after the last progress line). Runtime of the same job elsewhere: 50 min on the base branch (14269e0), 52 min on this PR's first head (b9f460c, green). The second commit adds ~25 in-memory unit tests that take seconds locally, so the suite is simply running at the edge of its budget and this attempt landed on a slower runner. Re-running the job once; if it times out again I'll look at where the time goes rather than re-run further.

---

yochze:
Correction to the above: I cannot re-run the job from here (the Actions re-run API returns 403 for this integration). The `unit-tests` job on 1170600 needs a manual "Re-run failed jobs" by a maintainer. Nothing in the PR changes what it runs; the budget is the issue. If the timeout keeps recurring across PRs, `timeout-minutes: 60` on the `unit-tests` job in `.github/workflows/e2e-tests.yml` is the knob (the base branch already takes 50 min there), but that is a workflow change outside this PR.

---

PR #1100, opened 2026-09-08
Refactor agent expansion to not auto-open Instructions group

yochze:
## Summary
Modified the agent expansion logic in KnowledgeExplorer to prevent automatically opening the Instructions group when an agent is expanded. The Instructions group now only opens on explicit user interaction.

## Key Changes
- Removed automatic expansion of the Instructions group (`'instr:' + id`) when an agent is opened
- Simplified the expansion condition to only call `loadAgentMeta()` without side effects on the Instructions group state
- Added clarifying comments explaining the new behavior: agent meta loads on expansion, but Instructions group remains collapsed until explicitly clicked or opened via `openAgentSection` or fresh create operations

## Implementation Details
The change preserves the agent metadata loading functionality while giving users more control over the UI state. The Instructions group will now only expand through:
- Explicit user clicks on the group
- Programmatic calls to `openAgentSection`
- Fresh agent creation flows

https://claude.ai/code/session_01984wzQmKv928fP1fiZ1dvE

### DIFF
--- a/frontend/components/KnowledgeExplorer.vue
+++ b/frontend/components/KnowledgeExplorer.vue
@@ -3292,7 +3292,9 @@ const expand = (key: string, force?: boolean) => {
   if (force) expanded.value.add(key)
   else if (expanded.value.has(key)) expanded.value.delete(key)
   else expanded.value.add(key)
-  if (key.startsWith('agent:') && expanded.value.has(key)) { const id = key.slice('agent:'.length); expanded.value.add('instr:' + id); loadAgentMeta(id) }
+  // Opening an agent loads its meta but leaves the Instructions group collapsed;
+  // it opens only on an explicit click (or openAgentSection / a fresh create).
+  if (key.startsWith('agent:') && expanded.value.has(key)) loadAgentMeta(key.slice('agent:'.length))
   // Lazy-load instruction rows on first expand of a group (rows arrive from the
   // backend; counts/badges were already loaded on mount).
   if (expanded.value.has(key)) {


yochze:
**CI: `e2e-tests (sqlite)` is red on a pre-existing intermittent failure, not this PR's.**

Failing test: `tests/e2e/test_sharepoint_onprem.py::test_blank_library_scope_is_identical_before_and_after_save` (`assert 0 == 3`).

Why it isn't this PR's:
- This PR changes one line in `frontend/components/KnowledgeExplorer.vue`; the backend e2e suite never loads it.
- The identical assertion fails on `main` too, in the two most recent non-cancelled failing runs: [run 34249645134](https://github.com/bagofwords1/bagofwords/actions/runs/34249645134) (merge of #1094) and [run 34188012062](https://github.com/bagofwords1/bagofwords/actions/runs/34188012062) (merge of #1085).

Likely root cause (for whoever picks it up): `POST /api/connections/{id}/refresh` submits the indexing job to the background daemon-thread loop in `connection_indexing_service.start` and returns immediately, but the test asserts `get_connection_tables(...)` right away without polling `GET /connections/{id}/indexing` for completion. Under `-n 2` xdist load the assert sometimes wins the race and sees zero tables. A fix would make the test poll the indexing row until it's finished before asserting.

No open PR carries a fix yet, so nothing to port. Re-running the failed job once.

---

yochze:
Follow-up: the re-run of `e2e-tests (sqlite)` was refused for this session (`403 Resource not accessible by integration`), so a maintainer needs to click **Re-run failed jobs** on [run 34258575618](https://github.com/bagofwords1/bagofwords/actions/runs/34258575618). Every other check on the head is green, including `e2e-tests (postgres)`, which runs the same test.

Proposed fix for the flake itself (kept out of this PR to avoid widening a frontend one-liner): `tests/e2e/test_connection_indexing.py` already has `_poll_until_terminal(...)`, which polls `GET /connections/{id}/indexing` until the job finishes. Reusing it in `tests/e2e/test_sharepoint_onprem.py` removes the race:

```python
from tests.e2e.test_connection_indexing import _poll_until_terminal

# in test_blank_library_scope_is_identical_before_and_after_save (and the
# lifecycle test above it, which has the same shape):
refresh_connection_schema(connection_id=conn["id"], user_token=token, org_id=org)
_poll_until_terminal(test_client, conn["id"], token, org, timeout_s=30.0)
assert len(get_connection_tables(connection_id=conn["id"], user_token=token, org_id=org)) == expected
```

---

PR #1112, opened 2026-09-12
Stop the agent losing context to a session race, and let the DB pool drain

yochze:
Investigated from a self-hosted Kubernetes + PostgreSQL report failing with
"remaining connection slots are reserved for roles with the SUPERUSER
attribute", alongside a complaint that the agent kept losing the thread of a
conversation. Both reproduced in a Postgres sandbox sized to the customer's
defaults (max_connections=100, 3 reserved); see
docs/feedback-loops/agent-context-loss-and-pool-drain-2026-09-12.md.

Context builders no longer race on the agent's session
------------------------------------------------------
ContextHub.refresh_warm and prime_static gathered their builders under
asyncio.gather(..., return_exceptions=True), but every builder is constructed
with the same AsyncSession. That is not safe for concurrent use, so the losers
raised "This session is provisioning a new connection; concurrent operations are
not permitted" — and each builder catches its own failure and returns an empty
section, so the planner was handed <queries> with no items as though the report
had none. A user refining a previous answer was talking to an agent that had
forgotten what it was refining.

The trigger is _release_db_between_steps(): it commits before each loop
iteration, leaving the session with no connection, which is exactly when the
provisioning window is widest. On a report with two saved queries the planner
saw zero of them in 10 of 10 refreshes following a commit; it now sees both in
10 of 10.

Serializing costs nothing — four builders on one session share one DBAPI
connection, so the gather was never parallel. Wall-clock is unchanged (14ms ->
13ms isolated; p50 149ms -> 170ms under 10-way concurrency, within noise), and a
builder that does fail is now logged with its section named.

The pool can now drain
----------------------
A QueuePool never shrinks: pool_recycle is applied lazily at checkout and
replaces a connection rather than closing it, and SQLAlchemy has no idle reaper.
Measured, a 20-connection pool stayed at 20 open backends indefinitely under
light traffic. Letting the server close idle sessions is what actually drains
it, with pool_pre_ping reconnecting transparently; pool_use_lifo then multiplies
the effect by concentrating traffic so the tail goes genuinely idle (settles at
4 open backends under FIFO, 1 under LIFO).

idle_session_timeout is PostgreSQL 14+ and an unrecognised GUC fails every
connect, so it is opt-in via BOW_DB_IDLE_SESSION_TIMEOUT_MS; the chart, which
bundles PostgreSQL 17, sets it. pool_size and max_overflow were hardcoded,
leaving an operator hitting slot exhaustion no lever short of editing the image
— they are now BOW_DB_POOL_SIZE / BOW_DB_MAX_OVERFLOW, and the chart sizes the
bundled server's max_connections to match what the app can actually ask for.

Two error messages stop saying the opposite of what happened
-----------------------------------------------------------
_mlog was a closure inside main_execution, referenced from two methods outside
that frame. Both sit in except blocks, so the NameError was caught and
relabelled as the failure of whatever the block guarded, on the success path:
focus-on-use logged "commit failed" after the commit had succeeded, and a
mid-run client that was built fine logged "construction failed".

Verified: an 8-agent concurrent run on the sandbox now completes with zero
errors of any class (was 26 context-section losses and 8 focus-on-use errors),
and idle connections after the run drop from 22 to 2.

Not done here, and deliberately: BOW_AGENT_SINGLE_WRITE_SESSION was benchmarked
for Postgres and is harmful — it corrupts the asyncpg connection ("got result
for unknown protocol state") across repeated runs, gives no connection saving
(peak 25 vs 24) and is slower. It stays opt-in. The chart also still sets no
limits.cpu, so worker count follows the node's core count rather than the CPU
request; changing that alters the resource contract of every deployment.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0121EQukxDsuZ6gAB3SGj4Q5

### DIFF
--- a/backend/app/ai/agent_v2.py
+++ b/backend/app/ai/agent_v2.py
@@ -1041,6 +1041,33 @@ class AgentV2:
         except Exception:
             return ()
 
+    def _mlog(self, label: str) -> None:
+        """Run-progress log line, timestamped against the start of the run.
+
+        This used to be a closure defined inside ``main_execution``, but two
+        call sites outside that frame referenced it — ``_persist_focus_on_use``
+        and ``_ensure_clients_for_attached`` — where the name does not resolve.
+        Both sit inside ``except Exception`` blocks, so the resulting
+        ``NameError: name '_mlog' is not defined`` was caught and relabelled as
+        the failure of whatever the block was guarding, on the *success* path:
+
+          - focus-on-use logged ``focus-on-use: commit failed`` after the commit
+            had already succeeded (8 times across a 10-agent run here), then ran
+            a no-op rollback;
+          - building a client for a mid-run attached agent logged
+            ``mid-run client construction failed`` for a client that was built
+            and registered fine.
+
+        Both messages point an operator at the opposite of what happened, and
+        they masked genuine failures of the same paths. Being a method keeps it
+        in scope everywhere.
+        """
+        import time as _time
+        t0 = getattr(self, "_run_t0", None)
+        rid = str(self.report_id)[:8] if self.report else "?"
+        elapsed = f" +{(_time.monotonic() - t0) * 1000:.0f}ms" if t0 else ""
+        logger.info(f"[agent:{rid}] {label}{elapsed}")
+
     async def _ensure_clients_for_attached(self) -> None:
         """Build query clients for agents attached AFTER run start (approved
         set_report_agents expansion) — without this create_data against the new
@@ -1060,7 +1087,7 @@ class AgentV2:
                     if built:
                         self.clients.update(built)
                         self.data_sources.append(ds)
-                        _mlog(f"mid-run client built for {ds.name}")
+                        self._mlog(f"mid-run client built for {ds.name}")
                 except Exception:
                     logger.exception("mid-run client construction failed for %s", getattr(ds, "name", "?"))
         except Exception:
@@ -1120,13 +1147,18 @@ class AgentV2:
             self.db.add(self.report)
             await self.db.commit()
             self._focus_set_by_use = True
-            _mlog(f"focus_on_use persisted={merged} via {tool_name}")
         except Exception:
             logger.exception("focus-on-use: commit failed")
             try:
                 await self.db.rollback()
             except Exception:
                 pass
+            return
+        # Outside the try on purpose: this block guards the commit, and anything
+        # raised by the success log would otherwise be reported as the commit
+        # having failed — which is exactly what happened while `_mlog` was an
+        # out-of-scope name here.
+        self._mlog(f"focus_on_use persisted={merged} via {tool_name}")
 
     def _resolve_instruction_scope_ids(self) -> Optional[List[str]]:
         """Data-source scope for the standing <instructions> block.
@@ -3983,10 +4015,9 @@ class AgentV2:
                 logger.warning("[headers] failed to rebuild planner LLM with membership identity", exc_info=True)
         try:
             import time as _time
-            _t0 = _time.monotonic()
-            _rid = str(self.report_id)[:8] if self.report else "?"
-            def _mlog(label):
-                logger.info(f"[agent:{_rid}] {label} +{(_time.monotonic()-_t0)*1000:.0f}ms")
+            # Anchor the run clock that self._mlog() reports elapsed against.
+            self._run_t0 = _time.monotonic()
+            _mlog = self._mlog
 
             # Start agent execution tracking
             self.current_execution = await self.project_manager.start_agent_execution(
--- a/backend/app/ai/context/context_hub.py
+++ b/backend/app/ai/context/context_hub.py
@@ -827,8 +827,14 @@ class ContextHub:
         previous answer ("now break that down by month") got an agent that had
         forgotten the thing it was refining.
 
-        Serializing costs nothing because there was no parallelism to lose, and
-        anything that still fails is logged at WARNING with the section named
+        Serializing costs nothing because there was no parallelism to lose:
+        benchmarked over three interleaved repetitions of 40 refreshes on a warm
+        session, p50 was 14.34 ms gathered versus 14.19 ms serialized. (On a
+        session that has just committed, the gathered version *is* a few ms
+        faster — because a builder that loses the race returns an empty section
+        without running its query at all.)
+
+        Anything that still fails is logged at WARNING with the section named
         instead of vanishing into a ``return_exceptions`` tuple. Returns results
         positionally, exceptions included, so callers keep their existing shape.
         """
--- a/backend/app/ai/context/context_hub.py
+++ b/backend/app/ai/context/context_hub.py
@@ -1,6 +1,7 @@
 """
 ContextHub - Main orchestrator for all agent context.
 """
+import asyncio
 import json
 import logging
 import time
@@ -290,6 +291,10 @@ class ContextHub:
         mode: Optional[str] = None,
     ):
         self.db = db
+        # Guards `self.db` across the context builders. Every builder here is
+        # constructed with this one AsyncSession, and an AsyncSession is not
+        # safe for concurrent use — see `_run_builders`.
+        self._db_lock = asyncio.Lock()
         self.organization = organization
         self.organization_settings = organization_settings
         self.data_sources = data_sources
@@ -699,7 +704,6 @@ class ContextHub:
             The user's query/prompt. If provided, enables intelligent instruction
             search to find relevant instructions beyond just 'always' load mode.
         """
-        import asyncio
         _t0 = time.monotonic()
 
         async def _timed(name, coro):
@@ -777,15 +781,16 @@ class ContextHub:
             _INSTRUCTIONS_CACHE[instr_key] = (time.monotonic(), built)
             return built
 
-        # Run static builders in parallel; schemas/instructions come from
-        # the cache when warm.
-        schemas, instructions, resources, files = await asyncio.gather(
-            _build_or_get_schemas(),
-            _build_or_get_instructions(),
-            _timed("resources", self.resource_builder.build()),
-            _timed("files", self.files_builder.build()),
-            return_exceptions=True,
-        )
+        # Static builders share self.db too, so they run serially for the same
+        # reason as the warm ones — see _run_builders. The stakes are higher
+        # here: a lost section is the schema or the instructions, and the agent
+        # would go on to plan against a connection it thinks has no tables.
+        schemas, instructions, resources, files = await self._run_builders("prime_static", [
+            ("schemas", _build_or_get_schemas()),
+            ("instructions", _build_or_get_instructions()),
+            ("resources", _timed("resources", self.resource_builder.build())),
+            ("files", _timed("files", self.files_builder.build())),
+        ])
         _hub_logger.info(f"[context_hub:prime_static] all_done +{(time.monotonic()-_t0)*1000:.0f}ms")
 
         # Store results (handle exceptions gracefully)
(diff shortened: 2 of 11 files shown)


yochze:
Reviewed `44bfa696228b2ae93d292ae573f4f41f2f0364b6`. I recommend addressing the following before merging; the cancellation item is lower priority.

1. **[P1] The Helm timeout default also applies to external PostgreSQL ≤13.** [values.yaml:143](https://github.com/bagofwords1/bagofwords/blob/44bfa696228b2ae93d292ae573f4f41f2f0364b6/k8s/chart/values.yaml#L143) defaults `idleSessionTimeoutMs` to `600000`, and the ConfigMap emits it even with `postgresql.enabled=false`. Existing external deployments upgrading with unchanged values inherit an unsupported startup parameter, causing all async connections to fail. The backend's opt-in compatibility protection is therefore bypassed by Helm. Enable the default only for the known bundled server, or keep external databases opt-in. Verified the external-database rendering; PostgreSQL 13 itself was not run locally.

2. **[P2] The new lock does not cover the complete concurrent context lifecycle.** [context_hub.py:841–853](https://github.com/bagofwords1/bagofwords/blob/44bfa696228b2ae93d292ae573f4f41f2f0364b6/backend/app/ai/context/context_hub.py#L841-L853) locks the builder list, but `prime_static()` executes `_instruction_query()` before acquiring it. `refresh_warm()` also reads organization settings before the lock and scheduled tasks after it. [Agent startup still gathers both methods](https://github.com/bagofwords1/bagofwords/blob/44bfa696228b2ae93d292ae573f4f41f2f0364b6/backend/app/ai/agent_v2.py#L4119-L4122). On a session without a checked-out connection, warm builders can acquire the lock while `_instruction_query()` is provisioning the connection, producing the same `InvalidRequestError` and an empty query section.

   Using the real hub, builders, ORM models, and SQLAlchemy session with two persisted queries, PostgreSQL 14 produced:

   | Lifecycle | Queries visible across five runs |
   |---|---|
   | Concurrent prime + warm, cold session | **0, 0, 0, 0, 0** |
   | Concurrent prime + warm, connected session | 2, 2, 2, 2, 2 |
   | Sequential prime + warm, cold session | 2, 2, 2, 2, 2 |
   | Standalone warm refresh, cold session | 2, 2, 2, 2, 2 |

   SQLite reproduced the concurrent-cold failure and sequential success as well. Comparing the same matrix with the pre-PR hub confirms that this is an **incomplete fix of the existing race**, not a newly introduced cold-initialization regression: standalone cold warm-refresh failed before and succeeds now. Normal agent initialization performs earlier database queries, so this probe does **not** establish that every ordinary startup has a cold session. Serialize the complete lifecycle calls, or consistently lock every shared-session access. Add a public-lifecycle regression test: the new tests only invoke `_run_builders()` and cannot detect this gap.

3. **[P2] Numeric `maxOverflow: 0` silently restores 20 overflow connections.** [config.yaml:83–88](https://github.com/bagofwords1/bagofwords/blob/44bfa696228b2ae93d292ae573f4f41f2f0364b6/k8s/chart/templates/config.yaml#L83-L88) tests values by truthiness. Rendering with `--set config.database.poolSize=5 --set config.database.maxOverflow=0` emits pool size 5 but omits `BOW_DB_MAX_OVERFLOW`; the backend then defaults to 20, permitting **25 connections per worker instead of five**. Quoted string `"0"` works, while numeric zero—explicitly allowed by the schema—does not. Numeric `poolSize: 0` is also omitted. Check for unset/empty-string values explicitly and add numeric-zero render coverage.

4. **[P3] Cancellation leaves unstarted coroutine objects unawaited.** All wrappers and builder coroutines are constructed before [the sequential loop](https://github.com/bagofwords1/bagofwords/blob/44bfa696228b2ae93d292ae573f4f41f2f0364b6/backend/app/ai/context/context_hub.py#L841-L853). Cancelling during the first await exits without awaiting or closing the remaining objects. An unchanged-helper probe produced three wrapper `RuntimeWarning: coroutine ... was never awaited` records plus an inner-builder warning during later event-loop cleanup. The lock releases correctly; this is cleanup/log noise, not evidence of a deadlock or persistent DB connection leak. Prefer coroutine factories instantiated as the loop advances, or explicitly close unstarted coroutines on cancellation.

**Validation:** all 13 new Python tests passed; `helm lint` passed. Direct Helm rendering reproduced both configuration findings. `helm unittest` was unavailable because that plugin is not installed. The actual production async-engine factory passed a PostgreSQL drain/reconnect probe: pool size 4, overflow 0, and a test-only 1000ms timeout opened four distinct backends; after 1.6 seconds idle there were zero, and the next checkout successfully ran `SELECT 42` and left one live backend. No full agent/LLM run, Kubernetes deployment, or production-scale load test was performed. Application code was not modified.

yochze:
All four addressed in `bfa0675`. All four reproduced first — thanks, #2 in particular was a real hole in the fix.

**1 (P1) — Helm default reached external databases.** Correct, and it reintroduced through Helm exactly the hard break the backend's opt-in default exists to avoid. `values.yaml` now leaves `idleSessionTimeoutMs` empty; the ConfigMap defaults it to `600000` only for the bundled subchart (PostgreSQL 17, a version we control) — conditioned on `postgresql.enabled` and an empty `database.host` — with an explicit setting always winning so an operator on an external 14+ server can still opt in. Verified across seven render scenarios:

| scenario | `BOW_DB_IDLE_SESSION_TIMEOUT_MS` |
|---|---|
| defaults (bundled) | `"600000"` |
| `database.host=rds.example.com` | *absent* |
| `postgresql.enabled=false` | *absent* |
| external + explicit `600000` | `"600000"` |

**2 (P2) — the lock did not cover the lifecycle.** Confirmed, and this was the important one: guarding only the builder list left the same race one layer up, since `prime_static` runs `_instruction_query()` before the guard, `refresh_warm` reads org settings before it and scheduled tasks after it, and `agent_v2.py:4119-4122` gathers both on one hub and one session. Each public method is now a thin locking wrapper over its existing body, so the guard spans everything that touches the session, and `_run_builders` raises if a caller has not taken the lock. Reproducing your matrix, queries visible to the planner over five runs:

| lifecycle | before | after |
|---|---|---|
| **concurrent prime + warm, cold session** | **0, 0, 0, 0, 0** | **2, 2, 2, 2, 2** |
| concurrent prime + warm, connected | 2, 2, 2, 2, 2 | 2, 2, 2, 2, 2 |
| sequential prime + warm, cold | 2, 2, 2, 2, 2 | 2, 2, 2, 2, 2 |
| standalone warm refresh, cold | 2, 2, 2, 2, 2 | 2, 2, 2, 2, 2 |

Your point that this is an incomplete fix of the existing race rather than a new regression matches what I see, and the caveat about ordinary startup not necessarily having a cold session is fair — the mechanism is real either way, so the guard belongs at the entry points regardless.

**3 (P2) — numeric zero dropped.** Confirmed: `--set config.database.poolSize=5 --set config.database.maxOverflow=0` emitted pool size 5 and no overflow var, so the backend applied its own default of 20 — 25 connections per worker where five were asked for. All three knobs now test for unset explicitly (`kindIs "invalid"` / empty-string) rather than truthiness, with render coverage for numeric zero, including `idleSessionTimeoutMs: 0` meaning "disable" rather than "use the default".

**4 (P3) — unstarted coroutines.** Confirmed. `_run_builders` now takes zero-argument factories and builds each coroutine when its turn comes, so cancellation leaves nothing behind.

**On the test gap you flagged:** you were right that the previous tests only invoked `_run_builders()` and could not detect this. Added a public-entry-point serialization test over `prime_static`/`refresh_warm`, a test that `_run_builders` refuses to run unlocked (so re-introducing the narrow guard fails loudly), and a cancellation test asserting no later factory is invoked.

**Validation:** 107 Python tests pass, `helm lint` passes, and 21 `helm unittest` cases pass — the plugin needs pinning to `0.5.1` on Helm 3.16 (newer releases use a `platformHooks` key that 3.16 rejects), which is likely why it was unavailable for you. Live 8-agent concurrent run on the Postgres sandbox: zero context-section losses, zero unlocked-call errors, zero never-awaited warnings.

One thing I did **not** touch: the `Exception terminating connection` / `unknown protocol state` errors that show up intermittently under load are rooted in `usage_policy_service.run_blocking` calling `asyncio.get_running_loop()` from a worker thread with no loop, closing a pooled connection there. It appears in sandbox logs predating this branch and that file is untouched here, so I left it alone rather than widen the PR — happy to open a separate issue if useful.

---

yochze:
Re-reviewed `bfa067508a0009d8a45b44065feec514dba05127`. The context race, numeric-zero handling, and cancellation cleanup findings are resolved. One compatibility gap remains:

**[P1] Secret-supplied external PostgreSQL URLs still inherit the idle timeout.**

[config.yaml:108–109](https://github.com/bagofwords1/bagofwords/blob/bfa067508a0009d8a45b44065feec514dba05127/k8s/chart/templates/config.yaml#L108-L109) assumes the app uses bundled PostgreSQL when `database.host` is empty and the subchart is enabled. But the [documented Secret installation](https://github.com/bagofwords1/bagofwords/blob/bfa067508a0009d8a45b44065feec514dba05127/k8s/README.md#L174-L205) supplies `BOW_DATABASE_URL` through a Secret, leaving both those values at their defaults:

```bash
helm template review ./k8s/chart \
  --set postgresql.auth.existingSecret=bowapp-secrets \
  --set config.secretRef=bowapp-secrets \
  --show-only templates/config.yaml
```

Verified output: no generated `BOW_DATABASE_URL`, but `BOW_DB_IDLE_SESSION_TIMEOUT_MS: "600000"` is still emitted. The deployment loads the Secret after the ConfigMap, so a Secret pointing at external PostgreSQL ≤13 supplies the external URL without removing the unsupported timeout. Deploying a bundled server does not prove the application uses it when its URL comes from a Secret.

**Requested change:** when the database URL is delegated to a Secret, keep the timeout unset unless explicitly configured. Add the Secret-based installation to the render coverage. This finding is based on verified rendering and configuration flow; PostgreSQL 13 itself was not run locally.

**Independent verification of the resolved items:**

- The real PostgreSQL 14 lifecycle probe preserved both saved queries in all five repetitions of each scenario: concurrent cold, concurrent connected, sequential cold, and standalone cold warm refresh. The former concurrent-cold failure is fixed.
- Numeric pool/overflow/timeout zeros render correctly.
- Builders are now created lazily, and the cancellation test passes.
- Both Python test groups passed: **16 tests**. Seven standard Helm render scenarios and `helm lint` passed.

No other actionable regressions found in the follow-up diff. No full agent/LLM run or Kubernetes deployment was performed during this re-review.

yochze:
Fixed in `a0429ec`. Reproduced your render exactly — `existingSecret` + `secretRef` emitted `BOW_DB_IDLE_SESSION_TIMEOUT_MS: "600000"` with no generated `BOW_DATABASE_URL`.

Chasing it down, the gap is a bit wider than the repro: `k8s/README.md` documents `BOW_DATABASE_URL` as a first-class Secret key, and `config.secretRef` is loaded after the ConfigMap via `envFrom` — so **either** Secret path can point the app anywhere while `postgresql.enabled` and `database.host` both still look bundled. `config.secretRef` alone was leaking too, not just the two together.

So rather than special-casing `existingSecret`, the condition now mirrors the URL-generating branch directly above it: password auth, subchart enabled, no external `database.host`, and neither Secret path in play. The default applies only where this chart generates the URL itself and nothing can override it.

| scenario | `BOW_DB_IDLE_SESSION_TIMEOUT_MS` |
|---|---|
| defaults (bundled) | `"600000"` |
| **`existingSecret` + `secretRef`** (your repro) | *absent* |
| `existingSecret` alone | *absent* |
| **`config.secretRef` alone** | *absent* |
| external `database.host` | *absent* |
| `postgresql.enabled=false` | *absent* |
| IAM provider (`aws_iam`) | *absent* |
| Secret install + explicit opt-in | `"600000"` |
| explicit `0` on bundled (disable) | `"0"` |

The IAM row was also unprotected in principle — `database.auth.provider` was never part of the test — so that is covered now too.

What settled the direction for me is the asymmetry you have been pointing at from the first round: missing the default costs a quiet worker some idle connections, while applying it to a server older than 14 fails every connection on upgrade. Erring toward unset is the cheap mistake, so I stopped trying to infer "probably bundled" and tied it to the one case the chart can actually prove.

Added chart coverage for both Secret paths and the IAM provider, and `k8s/README.md` now tells Secret installs how to turn the timeout on (it is the path where operators will most often want it and no longer get it by default).

**Validation:** nine render scenarios as above, 26 `helm unittest` cases, `helm lint`, and 16 Python tests — all passing. No application code changed in this round.

If it is useful, the `helm unittest` plugin works on Helm 3.16 when pinned: `helm plugin install https://github.com/helm-unittest/helm-unittest --version 0.5.1` (newer releases use a `platformHooks` key that 3.16 rejects).

---

PR #1113, opened 2026-09-12
Add Google Cloud Vertex AI provider support

yochze:
## Summary
This PR adds comprehensive support for Google Cloud Vertex AI as an LLM provider, enabling users to access three distinct model families (Anthropic Claude, Google Gemini, and third-party MaaS models) through a single unified provider configuration.

### Backend Implementation
- **New `vertex_auth.py` module**: Handles Google Cloud authentication (ADC and service account modes), endpoint derivation for regional/global Vertex locations, and credential caching to avoid repeated RSA key parsing
- **Model routing logic**: Added `_is_gemini_model_id()` and `_is_anthropic_model_id()` functions to route models to appropriate transports (Anthropic Messages API, google-genai, or OpenAI-compatible surface)
- **LLM client integration**: Extended `Anthropic`, `Google`, and `OpenAI` clients to support Vertex-specific configuration (base URLs, authentication flows, timeouts)
- **Credential handling**: Implemented secure separation of non-secret config (project_id, location, auth_mode) in `additional_config` from encrypted service account keys
- **Usage tracking**: Updated `ConsoleService` to correctly account for cache tokens across all Claude deployments (first-party API, Bedrock, Azure, Vertex)

### Frontend Components
- **Provider configuration UI**: Added Vertex-specific credential fields (Project ID, Location, Authentication mode toggle) in both onboarding and provider management modals
- **Model management**: Extended model card modal to support custom Vertex model IDs (matching `EDITABLE_MODEL_ID_PROVIDER_TYPES`)
- **Brand resolution**: Updated LLM brand detection to correctly identify Claude/Gemini models regardless of hosting provider
- **Localization**: Added English translations for Vertex-specific UI labels

### Testing & Validation
- **Comprehensive unit tests** (`test_vertex_provider.py`): 362 lines covering endpoint derivation, transport selection, credential resolution, token refresh, and auth flows
- **E2E tests** (`test_completion_vertex.py`): Real streaming completion tests and configuration persistence validation
- **Smoke test script** (`vertex_smoke.py`): Live integration test exercising the full inference path for each model family
- **Test fixtures**: Added `create_vertex_provider_and_models` fixture supporting both ADC and service account authentication modes

### Model Catalog
- Added preset Vertex models (Gemini 3.8 Flash as default, Gemini 3.1 Pro Preview) with accurate pricing and context window information
- Supports custom model IDs for third-party publishers (xAI Grok, Z.ai GLM, Meta Llama, DeepSeek, etc.)

1. **Endpoint derivation**: Implements correct regional endpoint patterns (`{region}-aiplatform.googleapis.com`) and special cases (global, us, eu multi-region endpoints), working around a bug in the pinned anthropic SDK (0.40.x) that incorrectly prefixes the global endpoint

2. **Credential caching**: Uses fingerprinting (SHA256 of key JSON) to cache parsed credentials per (auth_mode, project_id, key) tuple, avoiding expensive RSA key parsing on every LLM instantiation in busy organizations

3. **Third-party MaaS routing**: Automatically routes non-Anthropic/non-Gemini models to the global endpoint only, as Vertex restricts third-party publishers to global availability

4. **Token refresh flow**: Implements custom `GoogleBearerAuth` httpx auth flow that automatically refreshes expired Google OAuth tokens during request execution

5. **Timeout configuration**: Provides environment-overridable timeout profile (300s read, 10s connect) tuned for reasoning models' variable latency on Vertex's OpenAI-compatible surface

https://claude.ai/code/session_013GEopSXPN4VYwzwD8hH72T


yochze:
## CI: two red checks, neither caused by this PR

**Failing on `6da3e73`:** `integration-llms` and `playwright-tests`.

**Root cause for both: the OpenAI API account has no credits.**

```
openai.RateLimitError: 429 - 'You have no credits remaining. Add credits to
continue using the API.'   type: insufficient_quota, code: credit_balance_exhausted
```

- `integration-llms` — 9 failures, all the `[openai]` parametrization of `tests/integrations/llm_clients.py` (6 passed, 14 skipped). The Anthropic and Google parametrizations pass.
- `playwright-tests` — `tests/onboarding/onboarding-wizard.spec.ts:94`, the *"step 2: configure OpenAI LLM provider"* test, timing out waiting for `/successfully connected/i`. That step performs a real OpenAI connection test, so it fails for the same reason.

### Why this isn't the PR's

Both checks are already red on `main` at this PR's base commit `abc1c085` ([run 34703278773](https://github.com/bagofwords1/bagofwords/actions/runs/34707983907)), and the playwright failure there is identical down to the line and the counts — same spec, same `:94`, same `1 failed / 84 did not run / 3 passed (1.9m)`.

This PR does touch `openai_client.py` and `pages/onboarding/llm.vue`, so I checked that specifically rather than relying on the base-branch comparison alone:

- The failing `integration-llms` traces run through `openai_responses_client.py`, which this PR does not modify.
- `OpenAi.__init__` only constructs a custom `http_client` when `auth`, `timeout`, or `verify_ssl` is passed; every existing caller takes the unchanged path (covered by `test_openai_client_without_auth_keeps_sdk_default_http_client`).
- The onboarding edits are all gated behind `provider_type === 'vertex'`; the OpenAI branch of that form is untouched, and the new field-filter and `canTestConnection` clauses sit after the existing `openai` early-returns.

### What would fix it

Adding credits to the OpenAI account used by CI — a billing action, not a code change. There is no fix to port into this PR, and I am not going to skip, disable, or quarantine the tests to get green.

I have not re-run the jobs: a re-run cannot clear a billing block, and the base-branch evidence above is the stronger of the two available checks for "not this PR's".

`unit-tests`, `e2e-tests (sqlite)` and `e2e-tests (postgres)` are still in progress here. All three were also red on `main` at `abc1c085`, but I will compare each failure reason against the base rather than assume, and will report or fix accordingly.

---

PR #1119, opened 2026-09-12
Codex/connection management

yochze:


yochze:
### `playwright-tests` is red, but not because of this PR

**Failing check:** `playwright-tests` (the `e2e tests` workflow) on `42396ab` — [run 34746280910](https://github.com/bagofwords1/bagofwords/actions/runs/34746280910).

**What fails:**

```
[onboarding] › tests/onboarding/onboarding-wizard.spec.ts:59:3 › step 2: configure OpenAI LLM provider
  Locator: getByText(/successfully connected/i)
  Expected: visible   Received: <element(s) not found>   Timeout: 30000ms
  at tests/onboarding/onboarding-wizard.spec.ts:94:61
1 failed · 84 did not run · 3 passed
```

**Why it is not this PR's.** The identical failure — same spec, same line 94, same locator, same tallies (`1 failed · 84 did not run · 3 passed`) — occurs on `main` at `9fb34a0`, this PR's own base: [run 34717902345](https://github.com/bagofwords1/bagofwords/actions/runs/34717902345). The `e2e tests` workflow has been red on `main` for six consecutive runs (#1113, #1114, #1115, #1116, #1117, #1118). Nothing in this PR touches the onboarding wizard or the OpenAI provider path.

**Root cause, and why there is no fix to port.** That test fills the `OPENAI_API_KEY_TEST` secret into the provider form, clicks *Test Connection*, and waits for "Successfully connected" — a live call to OpenAI. The test is not skipped, so the secret *is* set; the connection test itself is what fails. That points at an expired or quota-exhausted key rather than anything in the code, so there is no code change to port into this PR. It needs the repository secret rotated.

**Worth flagging separately:** the failure aborts the run — `84 did not run`. So this one credential is currently masking the entire e2e suite on every branch, `main` included, and has been since at least #1113.

**Re-run:** not spent, and deliberately so. Playwright already retried this test twice in-run with identical results, and six `main` runs agree, so re-running would only confirm what the base branch already shows. Happy to trigger one if you'd rather have it on the record.

---

yochze:
### `unit-tests` is red: 12 of 14 are pre-existing on `main`, 2 are this PR's

**Failing check:** `unit-tests` on `42396ab` — [job 103694697239](https://github.com/bagofwords1/bagofwords/actions/runs/34746280910/job/103694697239). `14 failed, 4882 passed`.

`main`'s own `unit-tests` job was **cancelled** (not completed) on the last run, so CI gives no baseline. I ran the six affected files locally at three revisions instead:

| revision | result |
|---|---|
| `42396ab` (this PR's head) | 14 failed |
| `935d456` (`main`) | **12 failed** |
| `0a26e30` (this PR, before the three fix commits) | the other **2** failed |

So 12 are pre-existing repo-wide and none of the 14 come from the fix commits. The 12 on `main`: `test_sklearn_sandbox` (×4, `UnsafePythonError: Forbidden attribute access: '__name__'`), `test_focus_on_use` (×3, `'AgentV2' object has no attribute 'report_id'`), `test_read_query_observation_code` (×3, `'_FakeDB' object has no attribute 'scalar'`), `test_artifact_relationship_loading` (×2).

The remaining two are this PR's, both from the `@discovery_progress` work:

**1. `test_infor_olap_client::TestFailureClassification::test_manager_success_reports_worker_url` — a real duplicated round-trip, not a stale expectation.**

`xmla_base.XmlaClient.test_connection` now probes when a catalog is configured:

```python
if self.catalog:
    self._list_cubes(self.catalog)   # probe, result discarded
    catalogs = [self.catalog]
```

`InforOlapClient.test_connection` then calls `self._list_cubes(self.catalog)` again to count cubes. Instrumented, `_list_cubes` fires **twice with the same argument** on one `test_connection()`, so every Infor OLAP connection test now issues two identical XMLA round-trips. The test stubs two POSTs, so the third surfaces as `Reached the XMLA endpoint but discovery failed: Unexpected extra POST`. Adding a third stub makes it pass, confirming the count is the only difference.

Worth noting the probe's own contract: `progress.py` says the plumbing reports "from inside their existing iteration loops — never from new round-trips" and gives a lifecycle "without extra I/O".

Two ways to resolve, and it's a call about intent rather than a mechanical fix:
- **Dedupe** — have the base hand the cubes it already fetched to the subclass so the count reuses them. Keeps the stronger connectivity check, drops the second round-trip, and the test passes unchanged.
- **Accept the extra probe** as the cost of not trusting a configured name, and stub the third response in the test.

**2. `test_network_dir_incremental_index::TestProgressAndCancel::test_progress_ticks_per_file` — stale expectation.**

`discovery_progress` emits `reporter.phase('discovering_schema')` *before* invoking the wrapped `get_schemas`, so the first tick is now `('discovering_schema', None, 0, 0)` where the test asserts `('indexing files', None, 0, 3)`. That leading tick is the decorator's documented purpose ("a start/end lifecycle"), so the test just needs to account for it.

I've stopped short of pushing either. Both sit in behaviour this PR introduced deliberately, and #1 in particular has two defensible answers — say which you want and I'll make the change and verify it.

---

yochze:
### `e2e-tests (sqlite)` is red — also pre-existing on `main`

**This PR** (`42396ab`, [job 103694697212](https://github.com/bagofwords1/bagofwords/actions/runs/34746280910/job/103694697212)): `2 failed, 1454 passed`
**`main`** (`9fb34a0`, [job 103618325678](https://github.com/bagofwords1/bagofwords/actions/runs/34717902345/job/103618325678)): `2 failed, 1442 passed`

Same two tests, same assertions, same values on both:

- `tests/e2e/rbac/test_rbac_training_mode.py::test_agent_creator_can_train_agent_they_created` — `assert 200 == 403` on `_set_mode(..., 'training')`
- `tests/e2e/test_diagnosis_explorer.py::test_summary_line_is_computed_from_the_query` — the summary gained an `'unindexed': 0` key and `cost_usd` / `p50_ms` are now exact where the test uses `pytest.approx`

The 12-test delta in the pass count is this PR's own added tests, all passing. Nothing here is this PR's, so there is no fix to port.

`e2e-tests (postgres)` was still running when this was written. It fails on `main` too; if it comes back with these same two tests I won't add another comment for it — assume it's covered by this one unless I say otherwise.

Taken together with the two comments above, every red check on this PR is now accounted for: `playwright-tests` and `e2e-tests` are pre-existing on `main`, and of `unit-tests`' 14 failures, 12 are pre-existing and 2 are this PR's (awaiting a decision on the Infor OLAP one).

---

PR #1122, opened 2026-09-13
feat(netapp): add ONTAP root-cause diagnostic queries

yochze:
Add a native NetApp ONTAP connector for discovering diagnostic tables and querying RCA evidence.
BOW previously had no NetApp connection type; the integration now targets the frozen ONTAP 9.14.1 REST contract directly.

Storage investigators can connect an on-premises cluster, select relevant tables, and query inventory, topology, performance, events and additional diagnostic resources. Results preserve missing values, source identity and query provenance, and refuse incomplete totals. This is a beta implementation verified with a simulated API; exact 9.14.1P9/customer acceptance remains open. No related issue was supplied.

### Customer experience

| Before | After |
|---|---|
| No NetApp connection or diagnostic schema | Requested NetApp logo, connection form, 466 selectable diagnostic tables |
| Before | After |

Connection flow

<details><summary>Connection test and selectable tables</summary>

Connection test
Tables

</details>

```mermaid
flowchart LR
    Before[Unknown NetApp connection type]:::bad
    Registry[Registry and encrypted system credentials] --> Catalog[Frozen Swagger catalog: get_tables]
    Catalog --> Select[Select relevant schemas]
    Select --> Query[Validated JSON execute_query]
    Query --> REST[Bounded GET-only ONTAP REST]
    REST --> DF[DataFrame: nullable values and provenance]
    DF --> RCA[BOW code executor and query history]
    Query --> Reject[Reject mutations, scope changes and partial totals]
    classDef bad fill:#fee2e2,stroke:#dc2626
    classDef good fill:#dcfce7,stroke:#16a34a
    class Registry,Catalog,Select,Query,REST,DF,RCA good
```

### Reviewer decision

| Item | Assessment |
|---|---|
| Risk | Moderate: broad vendor schema coverage without a real appliance; deliberately beta and awaiting customer validation |
| Blast radius | New source, registry-generated form/icon, and an opt-in JSON query-capture condition in the shared executor; no migration or dependency change |
| Intended behavior | 35 curated REST resources, 429 diagnostic resources, constituent view and derived membership table; JSON queries returning pandas DataFrames; system credentials only |
| Verified | Connector tests, real BOW API/database schema indexing, standard integration harness against synthetic HTTP, 20 diagnostic HTTP queries, production build and browser connection/table discovery |
| NOT verified | Exact P9/local Swagger, customer permissions, hardware/FC evidence, all endpoint families on a real system, estate-scale performance, TLS against a customer CA, or a natural-language LLM prompt |
| First review | `netapp_ontap_client.py` query/pagination policy, `tools/netapp/generate_catalog.py` profiles/exclusions, shared executor opt-in, feedback-loop evidence |

### Evidence

| Check | Observed result |
|---|---|
| Connector/catalog unit contracts | 65 passed, 0 failed; includes all 465 REST-backed table request constructions |
| NetApp API/database end-to-end | 1 passed |
| Generic connection/data-source end-to-end | 11 passed |
| Standard connector integration | 1 passed; 21 unrelated connector cases deselected |
| Actual HTTP diagnostic queries | 20/20 passed against the expanded loopback simulator |
| Frontend production build | Passed, exit 0; warnings remain |
| Browser | Test connection → 466-table discovery → connection created → agent table selector |
| Lint / whitespace | Targeted Ruff and `git diff --check` passed |
| Shared executor regressions | 62 passed, 0 failed, 0 skipped: timeout, concurrency, cancellation and usage |
| Failures attributable to this PR | None remaining in completed required checks; defects found and fixed are listed below |
| Automated review | All six findings fixed in `2434eb4c8`; eight inline threads replied to with evidence and resolved |

### Context map

- `backend/app/data_sources/clients/netapp_ontap_client.py`: query contract, identity, retention, normalization, pagination and budgets.
- `backend/app/data_sources/clients/netapp/catalog.json` and `tools/netapp/generate_catalog.py`: 504 GET dispositions, 464 REST profiles and 40 explicit exclusions. Generated JSON is collapsed by GitHub; the human-readable endpoint inventory is in `docs/design/netapp-ontap-9141-endpoints.md`.
- `backend/app/schemas/{data_source_registry.py,data_sources/configs.py}`: source registration, auth and UI form.
- `backend/app/ai/code_execution/code_execution.py`: opt-in capture for structured queries, preserving NetApp evidence in query history.
- `backend/tests/{unit/test_netapp_ontap_client.py,e2e/test_netapp_ontap.py,integrations/ds_clients.py}`: contracts, service integration and customer/simulator hook.
- `tools/netapp/README.md`: runnable simulator and customer verification kit. `docs/feedback-loops/netapp-ontap-connector.md`: reproduce → fix → verify evidence and limits.
- `docs/design/netapp-ontap-9141-connector-plan.md` and MCP source review: design rationale and upstream references.
- Implementation/evidence commit: `cc87ce1ee`; verification record: `95dfe369c`. Follow-up: customer-run exact-version, role, hardware, protocol, history and reconciliation acceptance.

### Implementation and review detail

Use direct REST; MCP was studied as reference and is not a runtime dependency. Discovery loads the local catalog instead of crawling an appliance. The executor allows reviewed GET profiles and fields only, uses verified TLS by default, scopes parents explicitly, follows short/empty continuation pages, rejects redirects and changed scope, and enforces time/row/response budgets. Secret/content operations and action-like query controls are excluded. `limit` is intentionally a complete-result ceiling, not a silent truncation instruction.

Logical volumes and FlexGroup constituents are separate grains. Arrays remain structured; volume-to-aggregate membership is a separate relation. Large nullable integers remain exact. History-tier selection uses the age of the incident start. Metric status, duration and timestamp remain available; current topology cannot establish historical placement. Raw counters require definitions and timed samples; no unsupported rate formula or historical-capacity claim is added.

Self-introduced defects caught and fixed: optional `version.full` becoming `pd.NA` broke the connection response; singleton requests initially inherited collection controls; constituents lacked an independent view; the integration harness inferred the wrong class capitalization; dictionary queries needed explicit shared-executor capture. Tests cover these paths. An initial shared-regression invocation incorrectly disabled conftest for database-dependent usage tests; it was rerun with normal fixture/model loading. The first build attempt encountered the isolated Nuxt dev build lock; the successful builds ran after stopping that dev server.

No LLM was configured in the isolated app, so a natural-language investigation was not run. A deterministic RCA calculation did run through the real BOW code executor. The simulator implements representative routes and forced pagination, not complete ONTAP semantics. Public Swagger coverage and a successful connection are not proof that every customer investigation is supported. CLI-only diagnostics, excluded reads, external switch/host evidence and historical capacity/configuration remain explicit gaps.

No deployment, paid lab, AWS resources, MCP server or customer changes were made. Existing unrelated working-tree changes were left untouched in the original checkout.

### Review corrections — 2434eb4c8

All six review findings are addressed. Swagger path-item/operation parameters now merge with reference resolution and operation overrides, repairing the 19 missing parent contracts. Both generation and runtime reject malformed contracts before HTTP. NFS remains historical: its documented timestamp filter is retained independently of nested response timestamps, and nested time/quality fields survive projection. Requests honor all declared controls; missing cluster UUIDs fail explicitly; retrieval timestamps use UTC datetime dtype even in empty/derived frames; invalid integer parents fail before HTTP.

Observed red → green: 13 new regression failures on the reviewed code; seven additional generator/loader guard failures against temporarily restored original files. Final checks: 65 connector/catalog tests, 12 API/database tests, one standard integration test and 20/20 actual HTTP simulator queries passed. Targeted Ruff, deterministic regeneration and whitespace checks passed. All 465 REST-backed selectable tables are exercised at the HTTP boundary; this is structural coverage, not appliance certification for all 466 tables. Frontend code is unchanged in this follow-up; previous build/UI evidence is retained. Existing NetApp connections indexed before the correction should refresh their schemas.


tools/netapp/generate_catalog.py
```diff
@@ -0,0 +1,246 @@
...
+        if not cols and not reason:
+            reason = "No structured 200 response schema; specialist adapter required"
+        name = aliases.get(path) or "diag_" + re.sub("[^a-zA-Z0-9]+", "_", path).strip(
+            "_"
+        )
+        if name in resources:
+            raise ValueError("Table name collision: " + name)
+        params = {}
+        parents = []
+        for param in op.get("parameters", []):
```
yochze: **Blocking:** this reads only operation-level `parameters`, so 19 of the 466 advertised tables ship unusable.

Swagger 2.0 also permits path parameters at the **path-item** level (`item["parameters"]`), applying to every operation under that path — and ONTAP uses this. Those paths come out of the generator with `parents: []` while `path` still contains its `{...}` segments.

Confirmed against the shipped `catalog.json`: 19 resources have an unresolved placeholder with no declared parent, including

```
/network/fc/fabrics/{fabric.name}/switches
/network/fc/fabrics/{fabric.name}/zones
/network/fc/interfaces/{uuid}
/name-services/dns/{uuid}
/protocols/cifs/shadow-copies/{client_uuid}
```

FC fabric/switch/zone evidence is close to the center of a storage RCA, so this isn't a quiet corner of the catalog.

Fix is to merge both levels before the loop:

```python
for param in list(item.get("parameters", [])) + list(op.get("parameters", [])):
```

Worth pairing with a catalog invariant test — *no resource may retain a `{...}` segment after substituting its declared parents* — since that is the check that would have caught this at generation time rather than at a customer's appliance.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+        if not isinstance(spec, dict) or set(spec) - allowed:
+            raise OntapQueryError("InvalidQuery", "Unknown query controls.")
+        if spec.get("table") == "volume_aggregates":
+            return self._volume_aggregates(spec)
+        resource = self._resource(spec.get("table"))
+        limit = self._integer(spec.get("limit", self.max_rows), "limit", 1, self.max_rows)
+        parents = spec.get("parent", {})
+        if not isinstance(parents, dict) or set(parents) != set(resource["parents"]):
+            raise OntapQueryError("InvalidQuery", "Supply exactly the required parent parameters from get_schema.")
+        path = "/api" + resource["path"]
```
yochze, line 395: This is where the generator bug in `tools/netapp/generate_catalog.py:175` turns into a misleading diagnosis.

For the 19 resources whose `parents` is empty but whose `path` still holds a `{...}` segment, the equality check above (`set(parents) != set(resource["parents"])`) forces the caller to pass `{}`. This loop then has nothing to substitute, and the literal brace segment goes out on the wire — e.g. `GET /api/network/fc/interfaces/{uuid}`.

The 404 that comes back is mapped at `:469` to `"unsupported"` in `self._availability`, so the connector reports to the investigator that *the appliance does not support this endpoint*. The appliance is fine; the catalog is malformed. Silent-and-wrong is worse here than loud-and-wrong, because the operator has no way to tell the difference from the UI.

Two defenses worth having, independent of the generator fix:

1. Refuse to load a catalog whose resources contain unresolved placeholders, at `_catalog()` time.
2. After substitution, assert `"{" not in path` before issuing the request, and raise `InvalidQuery` rather than letting it reach HTTP — a malformed path should never be attributable to the appliance.

---

backend/tests/unit/test_netapp_ontap_client.py
```diff
@@ -0,0 +1,342 @@
...
+        {"table": "volumes", "filter": {"return_records": False}},
+        {"table": "volumes", "filter": {"is_constituent": True}},
+        {"table": "volumes", "limit": True},
+        {"table": "volume_metrics", "parent": {"volume.uuid": "../../cluster"}, "lookback": "1h"},
+        {
+            "table": "diag_protocols_cifs_domains_svm_uuid",
+            "parent": {"svm.uuid": "s"},
+            "filter": {"rediscover_trusts": True},
+        },
+        {"table": "diag_storage_luns_uuid", "parent": {"uuid": "l"}, "filter": {"data.offset": 0}},
```
yochze, line 111: This case passes for the wrong reason, which is why the catalog bug survived the suite.

`diag_storage_luns_uuid` declares its `uuid` parameter at the **operation** level, so the generator picks it up and the resource gets a correct `parents: ["uuid"]`. The 19 resources that declare path parameters at the path-item level never appear in any test, so nothing here exercises the broken shape.

Suggest adding a test over the catalog itself rather than another hand-picked resource — something like:

```python
def test_no_resource_has_unresolved_path_placeholders():
    for name, resource in _catalog()["resources"].items():
        path = resource["path"]
        for parent in resource["parents"]:
            path = path.replace("{" + parent + "}", "x")
        assert "{" not in path, f"{name} has unresolved placeholders: {resource['path']}"
```

That holds for all 464 resources and keeps holding as the catalog is regenerated against future ONTAP releases, which is the real risk surface for a frozen vendor spec.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+            seconds = int(match[1]) * {"m": 60, "h": 3600, "d": 86400, "w": 604800}[match[2]]
+            if seconds > 31536000:
+                raise OntapQueryError("UnavailableHistory", "Requested lookback exceeds one year.")
+            start, end = now - timedelta(seconds=seconds), now
+        else:
+            start = self._date(spec.get("start_time"))
+            end = self._date(spec.get("end_time"))
+        if start >= end or end > now + timedelta(seconds=30):
+            raise OntapQueryError("InvalidQuery", "Time range must be ordered and cannot extend into the future.")
+        field = "timestamp" if "timestamp" in resource["filters"] else "time" if "time" in resource["filters"] else None
```
yochze, line 346: One resource is unqueryable in both directions: `diag_protocols_nfs_services_svm_uuid_metrics`.

It is flagged `history: true` but exposes neither `timestamp` nor `time` in `filters`, so:

- without time bounds → `:312` raises `InvalidQuery` ("History queries require lookback or explicit UTC bounds")
- with time bounds → this line resolves `field` to `None` and `:331` raises `InvalidQuery` ("This endpoint does not expose a supported timestamp filter")

There is no third path. I checked the rest of the catalog and it is the only one in this state — every other `history` resource carries a usable filter.

Either drop it from the catalog or stop marking it `history`, depending on whether the endpoint genuinely serves time series. A generator-side assertion that `history` implies a `timestamp`/`time` filter would keep it from recurring.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+                    raise OntapQueryError("InvalidQuery", "Invalid sort field or direction.")
+            params["order_by"] = ",".join(order)
+        now = datetime.now(UTC)
+        time_params, _ = self._time_params(spec, resource, now)
+        if set(time_params) & set(params):
+            raise OntapQueryError("InvalidQuery", "Time filters cannot override time controls.")
+        params.update(time_params)
+        params["fields"] = ",".join(fields)
+        if resource["collection"]:
+            params["max_records"] = min(500, limit)
```
yochze: Asymmetry with the `order_by` handling just above: `order_by` checks `resource["controls"]` at `:433` before being sent, but `fields`, `max_records` and `return_timeout` go out unconditionally.

Three resources in the catalog declare empty `controls` and still receive all three. ONTAP tolerates the extra query args today, so this is latent rather than broken — but the catalog is the thing claiming to be authoritative about what each endpoint accepts, and the executor only half-honors it. Gating these on `resource["controls"]` the same way `order_by` is gated would make the contract mean one thing.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+            raise OntapQueryError("InvalidResponse", "Cluster identity is absent.")
+        version = frame.iloc[0].get("version.full")
+        generation = frame.iloc[0].get("version.generation")
+        major = frame.iloc[0].get("version.major")
+        if pd.isna(generation) or pd.isna(major) or generation != 9 or major != 14:
+            raise OntapQueryError(
+                "UnsupportedVersion",
+                "This connector requires the ONTAP 9.14 release family; verify the appliance version.",
+            )
+        self._cluster = str(frame.iloc[0].get("uuid", self._cluster))
```
yochze: The `self._cluster` fallback here is unreachable, and the failure mode is a poisoned provenance column.

`uuid` is always among the requested fields on the line above, so the DataFrame always has a `uuid` column — even when the value is null. `Series.get` therefore returns `pd.NA` rather than falling through to the default, and `str(pd.NA)` is the literal `"<NA>"`.

That value is then stamped into `_bow_cluster` on every row of every subsequent query (`:476`). For a column whose whole job is identifying which appliance the evidence came from, a silent `"<NA>"` is worse than an error — it survives into query history and looks like data.

```python
uuid = frame.iloc[0].get("uuid")
self._cluster = str(uuid) if not pd.isna(uuid) else self._cluster
```

Given `test_connection` already raises `InvalidResponse` when cluster identity is absent, refusing outright on a null uuid may be the better call.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+    "volume.uuid": "volumes",
+    "location.volume.uuid": "volumes",
+    "aggregate.uuid": "aggregates",
+    "lun.uuid": "luns",
+    "igroup.uuid": "igroups",
+}
+_META = {
+    "_bow_cluster": "str",
+    "_bow_connection": "str",
+    "_bow_retrieved_at": "datetime",
```
yochze, line 72: `_bow_retrieved_at` is advertised as `datetime` but delivered as a string.

The row builder sets `now.isoformat()` at `:478`, and the dtype cast loop at `:487` iterates `fields` only — the `_META` columns are never cast, so this arrives as object dtype holding an ISO string.

This matters more than a normal dtype nit because the agent writes pandas code against the advertised schema. Anything reaching for `.dt` — a resample, a time-bucketed rollup, an `.dt.floor('h')` on the retrieval time — raises `AttributeError` at execution.

Either cast it alongside the typed fields:

```python
frame["_bow_retrieved_at"] = pd.array([now] * len(rows), dtype="datetime64[ns, UTC]")
```

or declare it `"str"` in `_META` and let consumers parse. Casting is the better fit given the column is a real timestamp.

---

backend/app/data_sources/clients/netapp_ontap_client.py
```diff
@@ -0,0 +1,676 @@
...
+                self._availability[spec["table"]] = {
+                    401: "authentication_failed",
+                    403: "permission_denied",
+                    404: "unsupported",
+                }.get(e.status, e.code)
+                raise
+        rows = []
+        for record in records:
+            row = {k: self._extract(record, k) for k in fields}
+            row.update(parents)
```
yochze, line 505: Optional / nit — flagging it because the error it produces points at the wrong party.

`row.update(parents)` overwrites the extracted value with the caller-supplied string for the 29 (resource, parent) pairs where the parent name is also a typed field — e.g. `index` on `diag_cluster_sensors_node_uuid_index`, which the catalog types as `integer`.

In practice this is benign: the values agree, and I verified under the pinned `pandas>=2.3.3,<2.4` that `pd.array(["5"], dtype="Int64")` coerces cleanly, so valid queries are unaffected.

The rough edge is the invalid case. Parent validation at `:380` only checks the value is a nonempty path segment, so a non-numeric id for an integer column reaches the cast and surfaces as `InvalidResponse` — "Invalid value type for index" — blaming the appliance's response for what was a bad request. Checking the supplied parent against the declared column type at validation time would keep `InvalidQuery` and `InvalidResponse` meaning what they say.

---

yochze (commented):
Reviewed the ~1,700 lines of hand-written code (the 2 MB catalog and the screenshots aside). I read every production hunk, cross-checked the generated `catalog.json` against the executor's assumptions with scripted queries, and verified the dtype behavior against the pinned pandas rather than assuming it.

**The safety model holds up.** GET-only, redirect refusal, continuation-scope pinning, secret sanitization, error-body suppression, and `limit` as a complete-result ceiling rather than a silent truncation instruction — I tried to find a way through these and couldn't. Marking the two big JSONs `linguist-generated` in `.gitattributes` was the right call too.

Every defect I found is in the **contract between the generated catalog and the executor**, not in the executor's logic.

| # | Finding | Severity |
|---|---|---|
| 1 | `generate_catalog.py` ignores path-item-level Swagger params → 19 of 466 tables always 404, misreported as appliance-unsupported | **Blocking** |
| 2 | `diag_protocols_nfs_services_svm_uuid_metrics` is `history` with no time filter → unqueryable in both directions | Should fix |
| 3 | `fields`/`max_records`/`return_timeout` sent regardless of `resource["controls"]`, unlike `order_by` | Should fix |
| 4 | `_bow_cluster` fallback unreachable → literal `"<NA>"` in provenance on every row | Should fix |
| 5 | `_bow_retrieved_at` advertised as `datetime`, delivered as a string → agent `.dt` code raises | Should fix |
| 6 | Parent value overwrites typed column; bad parent surfaces as `InvalidResponse` not `InvalidQuery` | Optional |

Only #1 is worth blocking on. It's a correctness bug in generated data with a failure mode that actively misleads: the connector tells the investigator the appliance doesn't support FC fabric, switch, zone and interface reads, when in fact those rows were never constructed correctly. For a storage RCA tool, losing the FC endpoints silently is close to the worst available outcome.

The related point is that the PR headlines **466 selectable diagnostic tables**, and that number was never validated end-to-end — 19 of them cannot return a row. The catalog invariant test suggested inline (no resource may retain a `{...}` segment after substituting its declared parents) is the cheap, durable version of that check, and it keeps holding as the catalog is regenerated for future ONTAP releases.

Worth saying: the PR description is unusually candid about what wasn't verified, and the "self-introduced defects caught and fixed" section is a genuinely good signal. The gap here isn't diligence, it's that the suite tests hand-picked resources instead of asserting properties over the whole generated catalog — so a systematic generator bug had no way to show up.

Happy to push the generator fix plus the invariant test if useful.

---

tools/netapp/generate_catalog.py
yochze: Fixed in 2434eb4c8. The generator resolves parameter references and merges path-item plus operation parameters by (in, name), preserving operation-level overrides. Parents, filters and controls now all use that merged set. Regenerated all 464 REST profiles; every advertised REST-backed table passes the path-binding/request-construction checks. Actual HTTP simulator queries now also cover FC fabric switches/zones and FC interface detail.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8. Catalog loading rejects missing/extraneous parents and malformed templates as InvalidCatalog. Execution also independently refuses remaining braces before making HTTP requests. File-boundary regression tests prove malformed catalogs never reach HTTP; those tests failed against the reviewed code and pass with the fix.

backend/tests/unit/test_netapp_ontap_client.py
yochze: Added catalog-wide parent/time invariants and request construction across all 465 REST-backed selectable tables in 2434eb4c8 (464 profiles plus the constituent view). Added generator tests for inherited/referenced parameters, operation overrides and rejection of malformed contracts. This closes the representative-only coverage gap; it does not claim appliance validation for every table.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8 while preserving historical NFS support. The frozen Swagger explicitly describes historical metrics and declares a timestamp query parameter; response timestamps are nested under v3/v4/v41. The generator now retains that documented filter independently of response-field names and records nested timestamp/duration/status fields, which the executor preserves. Added a regression test and an actual HTTP simulator check requiring time bounds.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8. fields, max_records and return_timeout are each gated by the merged control declarations. The three control-free profiles are singleton resources, so the previous code sent fields to them; they now send no undeclared controls. All three are covered by actual HTTP simulator routes that reject extra query controls, plus the catalog-wide request check.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8. test_connection now rejects null, empty and whitespace-only cluster UUIDs with InvalidResponse before changing provenance. Parameterized regression cases cover each missing-identity form; no '<NA>' fallback is stored.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8. _bow_retrieved_at is now datetime64[ns, UTC] in normal and derived result frames, including empty results. Four regression cases exercise populated/empty volumes and membership rows and verify that .dt.floor('h') works.

backend/app/data_sources/clients/netapp_ontap_client.py
yochze: Fixed in 2434eb4c8. Parent validation checks both the path-parameter declaration and any corresponding integer response column (important for sensors, whose path index is declared string). Nonnumeric/out-of-range identifiers fail as InvalidQuery before HTTP. Parent-only identity columns retain the existing string representation; valid typed columns still normalize through the declared dtype.

PR #1123, opened 2026-09-13
Add dark mode support to report view and ag-grid components

yochze:
## Summary
This PR adds comprehensive dark mode styling to the report view page and ag-grid table components, ensuring proper contrast and visual hierarchy across light and dark themes.

### Report View (`frontend/pages/reports/[id]/index.vue`)
- **Border colors**: Added `dark:border-gray-700` to separator dividers and image borders for proper contrast in dark mode
- **Fork summary cards**: Updated amber-themed card with dark mode variants (`dark:border-amber-500/30`, `dark:bg-amber-500/10`)
- **Text colors**: Added `dark:text-gray-300` and `dark:text-gray-100` to various text elements for readability
- **Badge/pill styling**: Updated all informational badges (yellow, blue, green) with dark mode color pairs:
  - Light: `bg-{color}-50 text-{color}-{7|8}00`
  - Dark: `dark:bg-{color}-500/15 dark:text-{color}-300`
- **Scheduled prompt button**: Added `dark:bg-gray-800/40` for better visual separation
- **Instructions popover**: Added dark mode text color and updated load-mode badge styling with dark variants
- **Markdown content styling**: Added comprehensive `:global()` CSS rules for dark mode markdown rendering:
  - Code blocks: `#1f2937` background with `#e5e7eb` text
  - Inline code: `#374151` background
  - Links: `#93c5fd` with `#bfdbfe` hover state
  - Blockquotes: `#4b5563` border
  - Tables: `#111827` cells with `#374151` borders and `#1f2937` headers

### AG-Grid Component (`frontend/components/AgGridComponent.vue`)
- **Token inheritance**: Added `.ag-grid-themed .ag-grid` rules to re-inherit theme tokens (`--ag-background-color`, `--ag-foreground-color`, `--ag-border-color`, etc.) so grid cells respect app theme instead of balham's hardcoded palette
- **Header styling**: Implemented `color-mix()` to tint header background toward text color (darkens in light mode, lifts in dark mode) while staying within the card's hue family
- **Row/column borders**: Derived `--ag-row-border-color` and `--ag-header-column-separator-color` from inherited `--ag-border-color`
- **Header hover states**: Re-declared hover/moving background colors to resolve against header color instead of wrapper background

### Entity Detail Panel (`frontend/components/EntityDetailPanel.vue`)
- **Type badges**: Updated metric/query type badges with dark mode variants
- **Status badges**: Added dark mode styling for draft, suggested, and other status indicators
[...]

PR #1124, opened 2026-09-13
Release 0.0.559 — umbrella for #1119, #1120, #1122, #1123, #1125, #1126, #1127

yochze:
Umbrella branch that rolls up seven open PRs, bumps the version and writes the release notes. Merging this closes all seven.

## What's in it

| PR | Branch | Head | Merged in as |
|---|---|---|---|
| #1119 — Codex/connection management | `codex/connection-management` | `42396ab` | `f990860` |
| #1120 — HubSpot MCP server catalog preset | `ai/busy-gauss-mee8sg` | `a1a63be` | `dcff164` |
| #1122 — NetApp ONTAP root-cause diagnostic queries | `codex/netapp-rca-connector` | `2434eb4` | `f4a4312` |
| #1123 — Dark mode for report view and ag-grid | `ai/magical-darwin-7h4wh6` | `3b03e05` | `55e8160` |
| #1125 — Role-scoped code visibility | `claude/admin-role-visibility-4ljyew` | `2818314` | `61d7527` + `923a581` |
| #1126 — Brocade SAN root-cause investigations | `codex/brocade-rca-connector` | `429335a` | `cfc826f` |
| #1127 — Background judge scoring on the small model | `ai/brave-galileo-6pxw1h` | `d6ef1a8` | `5fdb726` |

Each branch was merged with `--no-ff`, unmodified — all seven PR head commits are ancestors of this branch, so GitHub marks every one as merged once this lands. #1125 is merged twice because it picked up two more commits after its first merge here; `923a581` brings its current head in.

## How to merge

**Merge commit, not squash or rebase.** A squash rewrites the seven branches' commits into one, which breaks the ancestry GitHub uses to auto-close them — they would stay open and need closing by hand.

## What merging them together broke — fixed in `54ff3b4`

One real collision, which is the point of doing this as an umbrella: #1119 adds a contract that every registered catalog's `get_schemas` must emit its first progress event — and so be cancellable — before touching credentials, network or driver state. #1122's NetApp client was written against `main` and predates that contract, so `test_every_registered_catalog_can_cancel_before_external_io` **passes on each branch alone** and fails here with `('netapp_ontap', 'AttributeError')`.

Fixed by decorating NetApp's `get_schemas` with `@discovery_progress` and reporting the table walk through `discovery_items`, so indexing shows bounded progress across its 466 tables like every other connector. Both helpers are no-ops without a callback, so an agent calling `get_tables()` is unaffected. Verified red → green: contract test plus the 62 NetApp client unit tests (79 passed), `tests/e2e/test_netapp_ontap.py` passed, `ruff` clean.

#1126 adds another connector and meets the same contract — it was checked explicitly and **passes** (17/17); its `get_schemas` already reports a phase before touching `self`, so it needed no change.

## Merge conflicts

Three, all "both sides added their own thing in the same place", resolved keeping both:

- `CHANGELOG.md` (`61d7527`) — #1125 had already bumped `VERSION` to 0.0.559 and opened its own `0.0.559` section. Resolution keeps every line, ordered by PR number. The `VERSION` bump itself auto-resolved: both sides made the identical `0.0.558 → 0.0.559` change. One correction folded in — #1125's changelog line was tagged `(#1122)` on its own branch; it describes #1125, so it is tagged `(#1125)` here.
- `backend/app/schemas/data_sources/configs.py` (`cfc826f`) — #1122 and #1126 each appended their config/credentials classes and then opened `__all__` with their own entries. Both class pairs and both export entries kept.
- `backend/tests/integrations/ds_clients.py` (`cfc826f`) — both added a `DATA_SOURCES` entry and a `ds_kwargs` branch, and the two `return {` blocks had collapsed into one. Split back into two complete branches.

Auto-merged but verified by hand afterwards, since more than one branch touches them:

- `backend/app/schemas/data_source_registry.py` and `frontend/components/DataSourceIcon.vue` (#1120 + #1122 + #1126) — all three registrations and all three icon entries present.
- `backend/app/ai/agent_v2.py` (#1125 + #1127) — the Judge is built with `small_model` in both scoring paths and `self.judge` is gone (#1127), while #1125's code-visibility hooks are intact.
- `locales/{en,es,he}.json` (#1119 + #1125) — all 10 locale files parse; all 106 keys from #1119 and both keys from #1125 present, none lost.

`alembic` resolves to a single head, `codeperm01` (#1125's backfill), chaining onto `overlayconn01`, which is already on `main`. No other branch in this set adds a migration.

## Release notes (CHANGELOG 0.0.559)

- Connection setup and management are now one flow with an explicit account scope; schema discovery reports bounded progress across connectors; the agent landing page shows which connections it can reach (#1119)
- Fixed delegated (`user_required`) connections: paginated schema reads warm the per-user catalog overlay, a failed post-login overlay sync no longer discards OBO credentials, and the OAuth `redirect_uri` honors `X-Forwarded-Host` (#1119)
- HubSpot's hosted CRM MCP server added as a one-click catalog preset (#1120)
- NetApp ONTAP connector added, shipping as beta — 466 selectable diagnostic tables over the ONTAP 9.14.1 REST contract, read-only and bounded (#1122)
- Dark mode covers the report view, AgGrid tables and the query panel (#1123)
- Roles can withhold the agent's generated code via two new permissions, `view_code` and `run_custom_code`, enforced by redacting the API payload rather than hiding UI (#1125)
- Brocade Fabric OS connector added, shipping as beta — 12 primary SAN diagnostic tables (156 more opt-in), read-only over an allowlisted bounded REST path (#1126)
- Fixed background judge scoring being billed to the organization's big default model instead of the small default the feature gate checks (#1127)

## Verification

Structure: all seven PR heads are ancestors of the branch, the union diff matches the seven branches' diffs, every overlapping registration and locale key survived the merge, alembic has a single head, and the conflict-resolved modules compile.

CI: every failing test on the previous head was traced to its origin rather than waved off — see the two comments below. Of 17 failures across `unit-tests` and `e2e-tests (sqlite)`, exactly one was created by merging these PRs together (the NetApp collision above, now fixed); the other 16 were pre-existing on `main` (14, reproduced locally on `935d456`) or owned by #1119 (2, reproduced on its branch alone). #1125's newer head also carries a fix for one of the `main`-level RBAC e2e failures. Note `unit-tests` is `continue-on-error: true` by design in `.github/workflows/e2e-tests.yml`.

https://claude.ai/code/session_01AFmMJAgjHuTfn9B4fy9nvT


yochze:
### `unit-tests` on `61d7527`: 15 failures, one of them this branch's

Pushed a fix for the one failure that only exists because these PRs are in the same tree. Reporting the other 14 rather than touching them — they are not this branch's, and the details are below.

**Fixed here — `54ff3b4`**

`test_discovery_progress_contract.py::test_every_registered_catalog_can_cancel_before_external_io`, failing with `('netapp_ontap', 'AttributeError')`.

#1119 adds a contract that every registered catalog's `get_schemas` must emit its first progress event — and so be cancellable — before it touches credentials, network or driver state. #1122's NetApp client was written against `main` and predates that contract, so the two only collide once both are in one tree: `get_schemas` went straight to `self` with no progress event, and the cancelling callback never ran.

This is the collision an umbrella is for. Verified locally that the test **passes on `codex/connection-management` alone** (17/17) and **fails on this branch** with exactly the one entry CI reports. The fix decorates `get_schemas` with `@discovery_progress` and reports the table walk through `discovery_items`, so indexing also shows bounded progress across the 466 tables instead of an open spinner — what #1119 gives every other connector. Both helpers are no-ops without a callback, so an agent calling `get_tables()` is unaffected.

Red → green after the fix: contract test plus the 62 NetApp client unit tests, 79 passed; `tests/e2e/test_netapp_ontap.py` passed; `ruff` clean.

**Not this branch's — 14 failures, no change made**

| Failures | Where they already fail |
|---|---|
| `test_artifact_relationship_loading` ×2, `test_focus_on_use` ×3, `test_read_query_observation_code` ×3, `test_sklearn_sandbox` ×4 | Pre-existing. These 12 are exactly the failure set of #1122's own run (`main` + the NetApp connector), and none of the tests touch NetApp. |
| `test_infor_olap_client::test_manager_success_reports_worker_url`, `test_network_dir_incremental_index::test_progress_ticks_per_file` | #1119's. Reproduced locally on `codex/connection-management` alone; both are its discovery-progress rework (`'discovering_schema'` vs the expected `'indexing files'`). |

That accounts for all 15: 12 pre-existing + 2 from #1119 + 1 from the merge, now fixed. #1125 added none — the run went from 4955 to 4995 passed with the same 15 failures.

No fix exists to port for the other 14, and fixing them here would mean rewriting #1119's tests or unrelated pre-existing ones inside a release branch. They belong on their own PRs. Worth noting `unit-tests` is `continue-on-error: true` by design in `.github/workflows/e2e-tests.yml`, so it is visible but non-blocking; the `e2e-tests` matrix is the gate and is still running on the new head.

No re-run spent — none of this is a flake.

---

yochze:
### `e2e-tests (sqlite)`: 2 failures, both already red on `main`

Standing down on these — they are not this branch's, and no fix exists to port.

```
FAILED tests/e2e/rbac/test_rbac_training_mode.py::test_agent_creator_can_train_agent_they_created
        AssertionError: assert 200 == 403
FAILED tests/e2e/test_diagnosis_explorer.py::test_summary_line_is_computed_from_the_query
        Left contains 1 more item: {'unindexed': 0}
```

Reproduced all three ways locally, same two failures, nothing else:

| Tree | Result |
|---|---|
| `61d7527` (this branch) | 2 failed |
| `codex/connection-management` alone (#1119) | 2 failed |
| **`origin/main` (`935d456`)** | **2 failed** |

Red on the base branch is the definition of not-this-PR's. Neither test file is touched by any of the five PRs (`git log origin/main..HEAD -- <both files>` is empty), and for the diagnosis one the cause is visible in `main` itself: `backend/app/services/diagnosis/service.py:405` already returns `"unindexed"` in the summary, while the test asserts an exact dict without it. The test and the service disagree on `main`; merging these five PRs changes nothing about that.

No re-run spent. A re-run is for confirming a suspected flake, and this is deterministic — three trees, three identical results. Fixing either one here would mean editing unrelated tests inside a release branch; they belong on their own PR against `main`.

**Where that leaves #1124:** every failing check on this branch is either pre-existing on `main` or owned by a contributing PR, except the one NetApp/discovery-contract collision, which is fixed in `54ff3b4`. Full CI is re-running on that head now and I'll report what comes back.

---

PR #1125, opened 2026-09-13
Add role-scoped code visibility with view_code and run_custom_code permissions

yochze:
## Summary

Implements role-scoped control over generated code visibility. Admins can now define roles that withhold the agent's generated SQL/Python code, with two new org-level permissions: `view_code` (see generated code) and `run_custom_code` (edit and execute custom code). The feature enforces visibility at the API level through server-side redaction, not just UI gating.

## Key Changes

**Core Permission System**
- Added `view_code` and `run_custom_code` to the permissions registry
- `run_custom_code` implies `view_code` (editing code you cannot read is incoherent)
- Both permissions default to ON for backward compatibility; withholding is the admin's deliberate act
- Added migration to backfill both permissions onto all existing roles

**Code Redaction Infrastructure** (`app/core/code_visibility.py`)
- Central redaction module that handles code visibility across all surfaces
- Request-scoped context variable tracks visibility decision per request
- Redacts code to `None` (not empty string) to distinguish from legitimately empty steps
- Covers three distinct surfaces:
  1. Step payloads (`StepSchema.code` / `PublicStepSchema.code`)
  2. Tool execution results (`result_json.code`, `.errors`, `.executed_queries`, `.error.failed_sql`)
  3. Per-query timing telemetry (`query_timings[].sql`, `sub_timings_json.queries[].sql`)

**API Enforcement**
- `POST /queries/{id}/run` and `POST /queries/{id}/preview` now require `run_custom_code` permission
- Step serializers apply code redaction based on resolved permissions
- Tool execution results redacted before serialization
- Published report path respects code visibility for unauthenticated users

**Prompt Constraint**
- System prompt includes code-visibility constraint when asker cannot view code
- Reduces unprompted code in model responses (though payload redaction is the actual enforcement)

**Removed Legacy Control**
- Deleted `enable_code_editing` org setting (was UI-only, never enforced server-side)
- Replaced by `run_custom_code` permission which is properly enforced

**Frontend Updates**
- Role editor now shows `view_code` and `run_custom_code` in permission list
- Code tab and edit buttons hidden when user lacks `view_code`
- Custom code execution gated on `canRunCustomCode` permission
- Updated permission implication logic to mirror backend

- **Default-deny context variable**: Unset visibility flag defaults to deny, so a forgotten surface fails loudly (authorized user sees missing code) rather than silently (unauthorized user sees code)
- **Comprehensive test coverage**: E2E tests verify both permissive (code shown) and restrictive (code withheld) paths; unit tests cover all redaction surfaces
- **Feedback loop validation**: Live sandbox testing caught a timing telemetry leak that code-only reading would have missed
- **Idempotent migration**: Backfill handles partial upgrades and re-runs safely

https://claude.ai/code/session_01SbNXLnzAiJvZcNWf7zSgMV


yochze:
## CI is red on failures that pre-date this branch

`unit-tests` and `e2e-tests` are failing, and none of the failures are caused by this PR. I verified each one by running it in a git worktree checked out at the base commit (`935d456`) and comparing against this branch — not by reading the diff.

### `unit-tests` — 12 failures, all identical on `main`

| Test file | Failures | Error |
|---|---|---|
| `test_artifact_relationship_loading.py` | 2 | `assert not True` |
| `test_focus_on_use.py` | 3 | `AttributeError: 'AgentV2' object has no attribute 'report_id'` |
| `test_read_query_observation_code.py` | 3 | `AttributeError: '_FakeDB' object has no attribute 'scalar'` |
| `test_sklearn_sandbox.py` | 4 | `UnsafePythonError: Forbidden attribute access: '__name__'` |

Running those four files at `935d456` and on this branch gives byte-identical results (10 failed / 15 passed, and 2 failed / 4 passed). The `AgentV2` ones are worth singling out because this PR does touch `agent_v2.py` — they fail the same way without any of that change.

### `e2e-tests` — 2 failures, both from one upstream commit

Both trace to `9f910b5` *"feat(training): replace execution listing with scoped BOW data tables"* (2026-09-08), which changed behaviour without updating the tests:

1. **`test_rbac_training_mode.py::test_agent_creator_can_train_agent_they_created`** — that commit opened the empty-report training case (an agent manager may train against BOW run history with no business source attached), so the expected 403 became a 200. **Fixed in this PR** (2818314) — the test now asserts the actual security property (the inaccessible agent must not attach to the report), which is stronger than inferring it from a status code that has since changed meaning.

2. **`test_diagnosis_explorer.py::test_summary_line_is_computed_from_the_query`** — the same commit added `unindexed` to the diagnosis summary; the test asserts exact dict equality and was never updated. **Not fixed here** — it is unrelated to code visibility and this branch has already absorbed two test corrections. One-line patch:

```python
assert s == {"matched": 2, "errors": 2, "users": 2, "unindexed": 0,
             "cost_usd": pytest.approx(0.55), "p50_ms": pytest.approx(42_000)}
```

### Two things worth acting on separately

- **`e2e-tests` and `unit-tests` have been red on `main` since at least 2026-09-08.** That job was the only thing that would have caught either regression above, and both landed in the same commit.
- **`test_rbac_complex_roles.py` never runs in CI.** Every test there calls `_requires_enterprise()` and `pytest.skip`s without a license, so all 25 are part of the 46 skipped. Nine of them were genuinely broken by the permission-string validation in this PR — I only saw that because my sandbox has enterprise features available, and fixed it in c9ccc73. Worth knowing that the custom-role paths, including role permission validation, are unverified by CI.

I have not re-run any job: reproducing the failures at the base commit is stronger evidence than a re-run, and a re-run would not change a deterministic assertion error.

---

PR #1133, opened 2026-09-14
feat(artifacts): verify interactive data apps with scoped browser tools

yochze:
Let the main agent verify interactive data apps through the existing browser tools, with an organization opt-in and one expandable verification group.
An initial render check cannot prove filters or backend updates work; verification previously also lacked a live organization/privacy gate.

## Reviewer decision

| | Decision context |
|---|---|
| Risk | Moderate: authenticated browser transport, live authorization, query/result correlation and session cleanup. |
| Blast radius | Five existing browser tools, main-agent guidance, artifact host diagnostics, transcript grouping and organization AI Settings. No new model-facing tools or DB migration. |
| Intended behavior | **Verify data apps defaults off.** Internal verification requires `allow_llm_see_data=true`, `enable_artifact_verification=true`, and an accessible saved page in the current report. Connector browsing requires privacy permission plus an authorized attached connector, independently of the new flag. |
| Verified | 167 focused backend tests passed (87 browser/artifact unit + 71 parameter + 9 policy/API cases); English/Hebrew settings capture and persisted off→on→off eligibility. Earlier commits in this PR include actual Luna complex/static generation and real browser/date checks. |
| Not verified | PostgreSQL, a fresh air-gapped container, future write/MCP operations, every generated app, or the full application suite. Natural-language generation was not rerun for the gating follow-up. |
| First review targets | `artifact_verification_policy.py`, `_browser_policy.py`, `artifact_preview_service.py`, and `ArtifactFrame.vue` request/revision correlation. |

## Customer impact

Users can see what the agent checked and inspect the supporting query/runtime evidence. Organization administrators choose whether to enable internal verification. Turning off data visibility blocks browser data access even if the agent already has an open session. Saved reports remain viewable normally.

| Before | After |
|---|---|
| A screenshot could look correct while filters, detail dialogs or query updates were broken. | The main agent exercises representative controls, inspects backend and visible results, and can repair/recheck the saved version. |
| Tool access depended on a deployment variable and settings captured at run start. | Organization policy is checked before each planner step, before browser operations and before evidence is returned. |
| Complex apps could receive an unavailable hint that still recommended browsing. | Hints recommend checks only when the exact saved target is eligible; static dashboards finish without interactive verification. |
| Checks and screenshots appeared separately. | Friendly changing titles, spinner and shimmer stay in one expandable group. |

Verification grouping

[Settings before](https://raw.githubusercontent.com/bagofwords1/bagofwords/codex/internal-artifact-verification/media/pr/internal-artifact-verification/settings-before.png) · [Settings after](https://raw.githubusercontent.com/bagofwords1/bagofwords/codex/internal-artifact-verification/media/pr/internal-artifact-verification/settings-after.png) · [Hebrew](https://raw.githubusercontent.com/bagofwords1/bagofwords/codex/internal-artifact-verification/media/pr/internal-artifact-verification/settings-he.png)

Organization verification toggle

## Mechanism

```mermaid
flowchart TD
  P[Before each planner step] --> G{Privacy + org opt-in + accessible page?}
  G -->|Yes| T[Expose existing browser tools]
  G -->|No| C{Privacy + authorized connector?}
  C -->|Yes| T
  C -->|No| H[Hide browser tools]
  T --> R[Recheck exact target before operation]
  R --> V[Authorized browser operation]
  V --> E[Snapshot + internal query/runtime evidence]
  E --> A{Access still allowed?}
  A -->|Yes| M[Main agent compares results / repairs / rechecks]
  A -->|No or unavailable| X[Close session and withhold evidence]
  B[Cached catalog alone] --> U[Revoked session still readable]
  classDef old fill:#fee2e2,stroke:#dc2626,color:#7f1d1d
  classDef good fill:#dcfce7,stroke:#16a34a,color:#14532d
  class B,U old
  class R,A,X,M good
```

`browser_navigate(artifact_id=...)` opens a saved page without a browser connector. A backend-only JWT broker permits scoped artifact reads, assets and declared viewer queries through normal authenticated APIs. The browser receives no token. Arbitrary writes/MCP calls are rejected.

`browser_act(expect_query_update={"params": {...}})` derives affected queries from the authorized manifest and returns a fresh snapshot plus compact evidence. Successful checks need no mechanical inspect/snapshot chain. Missing acknowledgements, parameter mismatches, runtime errors, ambiguous dates, empty date results and capped results remain visible findings or explicit inconclusive checks. Pending work is neutral until settled.

Live gating uses fresh DB sessions and the existing owner/admin/project/sharing permission rules. Saving the first page makes browser tools available on the next planner step; deletion or revocation removes eligibility. Mode/platform restrictions and unrelated dynamically registered tools are preserved. There is no deployment enable/disable variable.

## Evidence

| Check | Observed result |
|---|---|
| Latest organization gate/runtime tests | **9 passed**, including catalog transitions, non-admin revocation, opt-in/privacy gates, startup outages and logout. |
| Latest artifact/browser/date regression tests | **87 browser/artifact + 71 parameter tests passed**, including real Chromium cases. |
| Locale sweep | Incomplete: stopped after repeated localhost page failures (11 failed, 2 interrupted, 17 not run); no green claim. New settings labels were separately inspected in English/Hebrew and validated in all ten catalogs. |
| Settings UI/API | Off→on made the same report eligible; restoring off removed eligibility. English and Hebrew inspected; all ten catalogs add exactly the two expected keys with unchanged drift. |
| Red/green | Unavailable complex hints previously recommended navigation; corrected hints do not. Removing all five browser guards made both the denial check and mid-operation evidence-revocation check fail; restoring the guards passed. |
| Earlier Luna complex run | 8 browser calls, both backend queries updated, four-parameter combination, detail open/close, reset restored $2,328.60. |
| Earlier Luna static run | Static hint not recommended; zero browser calls. |
| Earlier frontend production build | Passed after grouping/delivery corrections. Not repeated for this schema/catalog-only UI addition. |
| Failures attributable to this follow-up at final validation | 0 observed in focused backend contracts. The locale sweep remains unverified because the local frontend became unreachable (connection refused), not attributed away as a proven baseline defect. |

The settings after/flow capture used the existing localhost frontend and real FastAPI handlers with the regular `app.db` through in-process ASGI transport while its existing reload worker was unresponsive. No second application server was started. No backend worker was killed or restarted during this follow-up. The localhost availability problem remains outside the completed gate tests.

During implementation the Hebrew check caught the new strings in the wrong namespace; that was corrected and recaptured. The prior date work also caught a saved-query compatibility regression, corrected in `aca7cee23`: extra metadata keys are ignored as before; individually valid mixed/reversed date bounds retain their order/precision. Strict calendar validation stays in the explicitly selected helper. No saved app/query is migrated or rewritten.

The natural-language weekly sales example still left capped totals insufficiently labeled in the app and crowded date controls at narrow widths. These remain generated-app quality findings; this PR does not claim every generated app is correct.

## Context map and review

| Area | Files / convention |
|---|---|
| Eligibility and runtime boundary | `backend/app/services/artifact_verification_policy.py`, `backend/app/ai/tools/implementations/_browser_policy.py`. New browser operations must use the shared pre-operation and pre-evidence guard. |
| Catalog and hints | `backend/app/ai/agent_v2.py`, `backend/app/ai/tools/artifact_verification.py`, `create_artifact.py`, `edit_artifact.py`. Refresh only browser descriptors each planner step; hints inspect the exact saved artifact. |
| Organization opt-in | `backend/app/schemas/organization_settings_schema.py`, `locales/*.json`. Existing settings API/UI; off by default. |
| Broker and evidence | `backend/app/services/artifact_preview_service.py`, `backend/app/ai/tools/implementations/_artifact_browser.py`, `frontend/components/dashboard/ArtifactFrame.vue`. Correlate accepted request IDs with runtime revisions. |
| Grouped UI | `frontend/composables/useBlockGrouping.ts`, `frontend/components/BlockGroupHeader.vue`, `frontend/components/tools/BrowserTool.vue`. |
| Date compatibility | `backend/app/ai/code_execution/query_params.py`; `docs/feedback-loops/date-range-backward-compatibility.md`. |
| Reproduction | `docs/feedback-loops/artifact-verification-organization-gate.md`, `docs/feedback-loops/internal-artifact-browser-verification.md`, `docs/feedback-loops/artifact-verification-review-fixes.md`. |

Earlier review corrections (`033cbbb2b`) address cursor replay, active-session capacity, outages versus denials, pending-state UI and current request-ID delivery. Reviewer confirmations and the two withdrawn Vue-i18n false positives remain on the PR. The later saved-date compatibility comment is addressed by `aca7cee23`; no claim is made that its author has re-reviewed it. Organization gating is implemented in `a70ec757d`.

Future live write/MCP verification needs a separate execution/approval contract; the current preview remains read-only. No linked issue.


backend/app/services/artifact_preview_service.py
```diff
@@ -0,0 +1,305 @@
...
+            known = set().union(*declared.values()) if declared else set()
+            if not names or not names <= known:
+                raise ValueError("Expected params must name declared input parameters")
+            ids = {qid for qid, fields in declared.items() if names & fields}
+        if not ids or not ids <= self.query_ids:
+            raise ValueError("Expected query IDs must belong to this artifact")
+        return {"query_ids": sorted(ids), "params": params}
+
+    async def wait(self, action_id=None, expectation=None, timeout=15):
+        since = self.actions.get(action_id, 0)
```
yochze: When `action_id` is `None`, `self.actions.get(None, 0)` yields `since = 0`, and the event filter below (`e["cursor"] > since and (not action_id or ...)`) degenerates to "every event since page load".

That path is taken twice: the initial navigate snapshot (`_artifact_browser.run_artifact_operation` → `operation == "snapshot" and not action_id`) and every bare `browser_snapshot` the model issues without `evidence_for_action_id`.

Consequence: `failed = any(e["kind"] == "error" ...)` sees the whole history, so a single load-time `pageerror` or console error pins `update_status` to `"failed"` for the rest of the session — every subsequent bare snapshot reports a failure that has nothing to do with what just happened. The frontend then paints those blocks amber via `verificationHasIssue`.

Suggest giving the no-action case a real floor rather than 0 — e.g. `since = self.actions.get(action_id, self.delivered)` if `action_id` is None, or recording a snapshot mark at the start of the call.

---

backend/app/ai/tools/implementations/_browser_common.py
```diff
@@ -302,53 +312,86 @@ async def _close(self, session_id: str):
...
             if existing:
                 existing.touch()
                 existing.patterns = patterns  # pick up config edits
                 existing.allow_downloads = allow_downloads
                 return existing
             if len(self._sessions) >= MAX_CONCURRENT_SESSIONS:
-                # Evict the least-recently-used to honor the cap.
-                lru = min(self._sessions.values(), key=lambda s: s.last_used)
-                await self._close(lru.session_id)
+                raise RuntimeError("Browser capacity is busy; finish without claiming interactive verification")
```
yochze: This replaces LRU eviction with a hard failure, which is a behavior regression for the existing browser-connector path, not just for internal previews.

`MAX_CONCURRENT_SESSIONS` is 3 and the idle TTL is 600s. This PR also makes connector sessions survive across turns (scope-keyed rather than closed per report), so live sessions now accumulate where they previously didn't. A fourth concurrent `browser_navigate` — an ordinary connector navigation for a different report/user, unrelated to artifact verification — now fails outright with "Browser capacity is busy; finish without claiming interactive verification", where before it evicted the LRU session and succeeded. The message is also verification-specific and will read as a non-sequitur in a plain browsing turn.

Suggest keeping LRU eviction for non-preview sessions (they're recreatable) and only refusing when every session at the cap is an in-flight preview, or raising the cap now that sessions are longer-lived.

---

backend/app/ai/tools/implementations/browser_navigate.py
```diff
@@ -56,7 +59,22 @@ def _fail(self, msg: str, code: str, **extra) -> ToolEndEvent:
...
+                result = await navigate_artifact(data, runtime_ctx)
+                yield ToolEndEvent(type="tool.end", payload=result)
+            except Exception as e:
+                yield self._fail(str(e)[:1500], "preview_unavailable")
+            return
+
+        if data.session_id:
+            prior = session_manager.get(data.session_id, runtime_ctx)
+            if prior is None or prior.preview:
+                yield self._fail("This session cannot be used for URL navigation", "session_scope_mismatch")
```
yochze, line 81: `session_manager.get()` returns `None` both for a genuine scope mismatch and for a session that simply no longer exists — idle-evicted past the 600s TTL, or closed by the new `close_execution` in `agent_v2`'s `finally`. Those two cases get the same hard failure here.

Previously `session_id` on navigate was effectively advisory: `open()` reused the report-keyed session or created a fresh one, so a stale id was harmless. The tool's own description (still "Open **(or reuse)** the report's browser session at a URL", and "Returns a session_id — pass it to ... to keep working in the same page") now contradicts the behavior: a model that follows the description and passes back its id after an eviction gets `session_scope_mismatch` instead of a new page.

Suggest failing only on an actual mismatch — i.e. when the session resolves but `prior.preview` is set (or its scope belongs to someone else) — and falling through to open a fresh session when the id is merely unknown.

---

backend/app/services/artifact_preview_service.py
```diff
@@ -0,0 +1,305 @@
...
+    def row_counts(data):
+        returned = len(data.get("rows") or [])
+        total = (data.get("info") or {}).get("total_rows")
+        return {"returned_rows": returned, "total_rows": total,
+                "result_truncated": isinstance(total, int) and total > returned}
+
+    async def _get(self, path):
+        response = await self.client.get(path)
+        if response.status_code != 200:
+            raise PermissionError(f"Preview access failed (HTTP {response.status_code})")
```
yochze: Every non-200 becomes a `PermissionError`, including 500/502/504 and timeouts-turned-error-status. That matters because of how the type is consumed downstream:

`run_artifact_operation` opens with a per-operation access recheck — `await preview._get(f"/api/artifacts/{preview.artifact['id']}")` — and `artifact_operation_failure` branches on exactly this type:

```python
output = {..., "error_code": "preview_restricted" if isinstance(exc, PermissionError) else "artifact_action_failed"}
if not isinstance(exc, PermissionError):
    ... evidence, snapshot, parameters ...
```

So a transient backend hiccup on that recheck is reported to the model as a permissions block ("Browser evidence is restricted by...") *and* discards the run's accumulated evidence, snapshot and parameter manifest — the exact data the deny-path is deliberately withholding for real denials. The model will then tell the user verification was restricted by policy when it wasn't.

Suggest raising `PermissionError` only for 401/403 and a distinct exception (or a plain `RuntimeError`) for everything else, so the non-denial branch keeps returning evidence.

---

frontend/components/BlockGroupTicker.vue
```diff
@@ -14,8 +16,8 @@
 				<span :key="shownLabel" :class="group.active ? 'ticker-shimmer' : ''">{{ shownLabel }}</span>
 			</Transition>
 		</span>
-		<span v-if="group.issueCount" class="text-amber-500">· {{ group.issueCount }} {{ group.issueCount === 1 ? 'issue' : 'issues' }}</span>
-	</div>
+		<span v-if="group.issueCount" class="text-amber-500">· {{ $t('tools.browser.checkIssues', { count: group.issueCount }) }}</span>
```
yochze, line 19: `tools.browser.checkIssues` is a pipe-plural message (`"{count} finding | {count} findings"`), but it's called without the plural-choice argument, so vue-i18n resolves to the first branch unconditionally — 3 findings renders as "3 finding". Same for `checkSummary` at line 49.

The repo convention passes the count as the third argument:

```
frontend/pages/projects/index.vue:195   $t('projects.reportCount', { count: project.report_count }, project.report_count)
frontend/components/ChangelogModal.vue:76   $t('changelog.changeCount', { count: v.entries.length }, v.entries.length)
```

Applies to all 10 locale files, several of which have more than two plural branches (ar, ru), so the choice arg is load-bearing beyond English.

```suggestion
		<span v-if="group.issueCount" class="text-amber-500">· {{ $t('tools.browser.checkIssues', { count: group.issueCount }, group.issueCount) }}</span>
```

---

frontend/components/BlockGroupTicker.vue
```diff
@@ -34,13 +36,17 @@
...
 const props = defineProps<{ group: BlockGroup; expanded: boolean }>()
 defineEmits<{ (e: 'toggle'): void }>()
 
 // Minimum time a label stays on screen before the next one may swap in.
 const MIN_HOLD_MS = 500
 
 function labelFor(g: BlockGroup): string {
+	if (g.verification) return g.active
+		? (g.runningLabel || t('tools.browser.verifying'))
+		: t('tools.browser.checkSummary', { count: g.count, title: g.lastTitle || t('tools.browser.interactionChecks') })
```
yochze, line 50: Same plural issue as line 19 — `checkSummary` is `"{count} check · {title} | {count} checks · {title}"` and needs the count as the third argument:

```suggestion
		: t('tools.browser.checkSummary', { count: g.count, title: g.lastTitle || t('tools.browser.interactionChecks') }, g.count)
```

---

frontend/composables/useBlockGrouping.ts
```diff
@@ -334,6 +336,18 @@ function buildGroup(run: any[]): BlockGroup {
...
+/** Internal verification is an explicit server-owned run, not generic browsing. */
+function verificationKey(block: any): string {
+  const te = block?.tool_execution
+  return te?.arguments_json?._verification_group_id || te?.result_json?.verification_group_id ||
+    (te?.tool_name === 'browser_navigate' && te?.arguments_json?.artifact_id ? 'opening-preview' : '')
+}
+function verificationHasIssue(block: any): boolean {
+  if (!verificationKey(block)) return false
+  const e = block?.tool_execution?.result_json?.evidence
+  return !!e?.errors?.length || ['failed', 'parameter_mismatch', 'data_not_acknowledged', 'no_expected_request', 'pending'].includes(e?.update_status)
```
yochze: `'pending'` in this list contradicts the policy this PR adds. `ARTIFACT_VERIFICATION_POLICY` in `backend/app/ai/tools/artifact_verification.py` states verbatim:

> A pending result is not a failure or pass.

and the locale string is the neutral `"pending": "Waiting for data"`. But including it here increments `issueCount`, so the group header renders amber with "N findings" for a state the backend explicitly declines to call either way. Worth dropping `'pending'` (and deciding whether `data_not_acknowledged` belongs, since it's the same 15s-timeout family) so the UI and the policy text agree.

---

frontend/components/dashboard/ArtifactFrame.vue
```diff
@@ -556,8 +557,32 @@ const props = defineProps<{
...
+  try {
+    const value = await $fetch(path, options);
+    return { data: ref(value), error: ref(null), pending: ref(false), status: ref('success'), refresh: () => {} };
+  } catch (error) {
+    return { data: ref(null), error: ref(error), pending: ref(false), status: ref('error'), refresh: () => {} };
+  }
+}) as typeof useApplicationFetch;
+
+let verificationRevision = 0;
+const verificationRequests = new Set<string>();
```
yochze: Minor: this set is only ever added to (line 1229) and never cleared, so every `data_sent` event (line 2069) carries every request ID observed since mount, and `ArtifactPreviewService.record` stores that growing set under each revision.

Two consequences, both small at the current 40-query budget but worth a bound:

1. `wait()`'s `acknowledged` union gets monotonically more permissive — a `data_received` for any revision vouches for all historical request IDs, which weakens the correlation the mechanism exists to provide.
2. `record()` truncates at 8000 chars of serialized JSON and *replaces the whole payload* on overflow. If `request_ids` ever pushes a `data_sent` past that, the event becomes `{"message": "Diagnostic exceeded size limit", "truncated": true}` and the branch below stores `self.revisions[None] = set()` — after which `received` can never be satisfied and every action reports `data_not_acknowledged`.

Clearing the set in `sendDataToIframe` after emitting (or keeping only IDs since the last `data_received`) would fix both.

---

backend/app/ai/agent_v2.py
```diff
@@ -819,7 +823,7 @@ def _is_agent_file(f) -> bool:
         allow_llm_see_data_cfg = self.organization_settings.get_config("allow_llm_see_data") if self.organization_settings else None
         allow_llm_see_data = getattr(allow_llm_see_data_cfg, "value", True) if allow_llm_see_data_cfg is not None else True
         if not allow_llm_see_data:
-            all_catalog_dicts = [t for t in all_catalog_dicts if t['name'] != 'inspect_data']
+            all_catalog_dicts = [t for t in all_catalog_dicts if t['name'] != 'inspect_data' and not t['name'].startswith('browser_')]
```
yochze: Question on intended scope: this now hides *all* `browser_*` tools when `allow_llm_see_data` is off, not just the new internal-preview path. Orgs that have that setting disabled today keep their browser connector but silently lose every browser tool from the catalog after this merges.

There's already a narrower guard for the preview itself — `ArtifactPreviewService.open` and `run_artifact_operation` both raise `PermissionError` on the same config — so the internal path is covered without touching connector browsing. If the intent is broader (screenshots of connector pages can expose data too), that's defensible, but it's a customer-visible capability removal that isn't called out in the PR description and probably deserves its own line in the changelog.

---

yochze (commented):
Reviewed the code under `backend/app` and `frontend/` (skipped docs/media/log fixtures). The transport design holds up well — strict route allowlist, no JWT in the browser, WebSockets closed, `service_workers: "block"`, viewer-mode-only POSTs with a body-key whitelist, and a query budget. I couldn't find a way out of the broker scope.

Nine inline comments. The ones I'd want addressed before merge:

1. **`wait()` scans from cursor 0 when `action_id` is None** (`artifact_preview_service.py:242`) — one load-time error pins every later bare snapshot to `"failed"`.
2. **Session-cap eviction became a hard error** (`_browser_common.py:360`) — regresses ordinary connector browsing at the 4th concurrent session, now that sessions survive across turns.
3. **Unresolvable `session_id` hard-fails** (`browser_navigate.py:76`) — contradicts the tool's own "open or reuse" description after an idle eviction.
4. **Every non-200 is a `PermissionError`** (`artifact_preview_service.py:95`) — a transient 5xx on the per-operation recheck is reported as a policy denial *and* discards the run's evidence.

Then two UI correctness nits (pipe-plural i18n keys called without the plural arg; `'pending'` counted as a finding against this PR's own policy text), one minor unbounded-set issue, and one scope question about `browser_*` being dropped wholesale under `allow_llm_see_data`.

Things I checked and cleared: the broker's endpoint allowlist covers every `useMyFetch` path `ArtifactFrame` actually calls; `/api/api/` normalization is safe because it feeds an allowlist; the `expose_binding` receiver correctly rejects non-main-frame sources; the unauthenticated `/artifact-preview/` route carries no session and its API calls fail normally for an anonymous visitor; the `useMyFetch` shim is safe since all call sites await; the `onMounted` early return skips nothing but the background rerun; `row_counts` genuinely supports `result_truncated`; Playwright route ordering puts `preview.route` ahead of the allowlist guard. The two `test_read_query_loads_only_requested_query_graph` failures do reproduce on the merge base — the stale-test classification checks out.

---

backend/app/services/artifact_preview_service.py
yochze: Verified fixed in 033cbbb2b.

`since = self.actions[action_id] if action_id else (self.delivered if since is None else since)` gives bare snapshots a real floor at the delivery cursor, and `run_artifact_operation` now threads `since_cursor` through for the explicit case. A load-time error is delivered once and no longer re-reported by every later snapshot, while explicit action checks keep their full history. Resolving.

---

backend/app/ai/tools/implementations/_browser_common.py
yochze: Verified fixed in 033cbbb2b.

LRU reclaim is back for sessions that are genuinely idle (`not s.preview and not s.active_execution_ids`), and the hard error now fires only when every session at the cap is in use — with a message that matches that situation. `active_execution_ids` tracked on `get`/`open` and released in `close_execution` is a cleaner protection than the TTL alone, and `_evict_idle` correctly skips in-use sessions rather than pulling a page out from under a live turn. Resolving.

---

backend/app/ai/tools/implementations/browser_navigate.py
yochze: Verified fixed in 033cbbb2b.

`get(..., strict=True)` raises only when the session actually resolves and its scope doesn't match; an unknown id still returns `None` (the `if s and ...` guard never fires), so the `prior is None` path now falls through to `session_manager.open` and gets a fresh authorized session. Live foreign sessions and internal previews are still rejected. Behavior matches the tool description again. Resolving.

---

backend/app/services/artifact_preview_service.py
yochze: Verified fixed in 033cbbb2b.

401/403/404 stay `PermissionError`; every other non-200 and `httpx.RequestError` now raise `PreviewUnavailableError`, which `artifact_operation_failure` maps to a distinct `preview_unavailable` code, and `navigate_artifact` routes through the same path on startup. The misreporting I was worried about — a transient 5xx presented to the model as a policy block — is gone.

Noting for the record that you went narrower than my suggestion: evidence is still withheld on `PreviewUnavailableError`, not just on denial. That's a defensible call (an incomplete recheck can't confirm access is still valid) and it's stated in the description, so I'm not pushing back. Resolving.

---

frontend/components/BlockGroupTicker.vue
yochze: You're right and I was wrong — withdrawing this.

I installed the pinned major (vue-i18n 9.14.5) and ran it rather than arguing from the docs:

```
named only, 1 : "1 finding"
named only, 3 : "3 findings"
named + plural : "3 findings"
summary named  : "10 checks · X"
```

vue-i18n 9 resolves the plural index from a named `count`, so the third argument is redundant here and the existing call is correct. The sites I cited as "the convention" are just being explicit, not compensating for anything. My apologies for the noise — no change needed. Resolving.

---

frontend/components/BlockGroupTicker.vue
yochze: Same — false positive, withdrawing. Verified against vue-i18n 9.14.5 that `t('checkSummary', { count: 10, title: 'X' })` returns `"10 checks · X"`; the named `count` drives plural selection on its own. No change needed. Resolving.

---

frontend/composables/useBlockGrouping.ts
yochze: Verified fixed in 033cbbb2b.

`'pending'` is out of the issue list in both `verificationHasIssue` and `BrowserTool.vue`'s `iconColor`, and now has its own neutral channel: `pendingCount` tracked per action in `buildGroup`, a gray "Waiting for data" chip on the header, and a clock icon instead of amber. The per-action set/delete means a pending check that later settles clears itself rather than sticking. UI and `ARTIFACT_VERIFICATION_POLICY` agree now.

`data_not_acknowledged` staying a finding is the right call — that one is a concluded 15s timeout with no delivery, not an open question. Resolving.

---

frontend/components/dashboard/ArtifactFrame.vue
yochze: Verified fixed in 033cbbb2b, and the fix is better than what I suggested.

`Map<queryId, requestId>` keyed per query means a resend carries the *current* accepted result for each dataset rather than a growing history, `verificationRequests.clear()` in `fetchData` drops parameter-run IDs when fresh snapshots replace them, and building `request_ids` from `visualizationsData` scopes the payload to what was actually sent. Both concerns are gone: the acknowledgement union no longer widens over time, and the payload can't grow into `record()`'s 8000-char truncation. The `artifactVerificationDelivery.mjs` regression (`['1','2']` before → `['2']` after) pins it. Resolving.

---

backend/app/ai/agent_v2.py
yochze: Answered — that's all I was after. The intent is now stated in the description ("Disabling `allow_llm_see_data` also removes existing connector browser tools from the agent catalog because page text/screenshots expose data") and in the review-corrections section, so the capability removal is documented rather than silent. The reasoning holds: an accessibility snapshot or screenshot of a connector page is data exposure by the same measure `inspect_data` is. Resolving.

---

backend/app/ai/code_execution/query_params.py
```diff
@@ -97,10 +155,18 @@ def coerce_param_value(spec: ParamSpec, value: Any) -> Any:
...
         out = {}
         for k in ("from", "to"):
             if value.get(k) is not None:
                 out[k] = _coerce_scalar(ParamSpec(name=spec.name, type="date"), value[k])
+        if "from" in out and "to" in out:
+            start, end = (_temporal_key(out[k]) for k in ("from", "to"))
+            if start[0] != end[0]:
+                raise ParamError(f"param '{spec.name}': range bounds must use the same date/timestamp and timezone convention")
+            if start[1:] > end[1:]:
+                raise ParamError(f"param '{spec.name}': range 'from' must not be after 'to'")
```
yochze: Compatibility question on the three new `date_range` rejections, since this isn't a verification-only path.

`coerce_param_value` is reached from `resolve_param_values` → `query_service._resolve_params_for_run`, so it runs for **every** viewer parameter run on every already-saved query, not just for apps generated after this merges. Three inputs that previously resolved now raise `ParamError`:

1. **Extra keys** (159) — `{from, to, preset: 'last30'}` was silently narrowed to `from`/`to` before; it's now a hard error. Any already-saved generated app that stashes a preset or label alongside the bounds breaks on its next run.
2. **Mixed conventions** (167) — `{from: '2025-01-01', to: '2025-01-31T23:59:59Z'}` is a plausible thing for existing generated code to send (date-only lower bound, end-of-day upper) and is now rejected outright.
3. **Inverted range** (169) — previously returned zero rows; now errors. This one can also fire transiently from a two-input date picker mid-edit, where an empty result is arguably the friendlier intermediate state.

All three are defensible as *forward* contract tightening — the new `check_date_range_code` guard and the `create_data` prompt rules cover newly generated code well. The question is whether previously-saved queries were checked against them. If there's a migration/backfill concern here it's invisible in the diff, and if the answer is "generated apps only ever send `{from,to}` calendar dates", saying so explicitly would close it.

No concern with `calendar_date_bounds` itself — I confirmed `ParamSpec.required` defaults to `False`, so `calendar_date_bounds(None)` returns `(None, None)` rather than raising, which is the common null-default case the prompt instructs models to rely on.

---

yochze (commented):
Re-reviewed at 033cbbb2b. **All nine threads answered and resolved.**

Seven were real and are fixed — I verified each against the code rather than the description:

| Finding | Status |
|---|---|
| `wait()` scanning from cursor 0 | Fixed; bare snapshots floor at `self.delivered`, explicit checks keep history |
| Capacity hard-fail | Fixed; idle LRU reclaim restored, `active_execution_ids` protects live turns |
| Stale `session_id` | Fixed; `strict=True` raises only on a real mismatch, unknown ids open fresh |
| Non-200 → `PermissionError` | Fixed; `PreviewUnavailableError` + distinct `preview_unavailable` code |
| `'pending'` counted as a finding | Fixed; separate neutral `pendingCount`, gray clock, clears when the action settles |
| `verificationRequests` unbounded | Fixed, better than suggested — per-query `Map`, cleared on `fetchData` |
| `browser_*` under `allow_llm_see_data` | Answered; intent now documented in the description |

**Two were my mistake.** The plural-argument comments were wrong. I installed the pinned major and ran it instead of arguing from the docs — vue-i18n 9.14.5 resolves the plural index from a named `count`, so `t('checkIssues', { count: 3 })` correctly returns `"3 findings"` and the third argument is redundant. Your dismissal was right; sorry for the noise.

One new comment on 87e1ac61b, and it's a question rather than a defect: the `date_range` tightening in `coerce_param_value` lands on the shared `resolve_param_values` path, so it applies to every already-saved query, not only to newly generated code. Worth confirming that no existing generated app sends extra keys, mixed date/timestamp bounds, or a transiently inverted range. I checked `calendar_date_bounds(None) → (None, None)` and that's fine.

Separately, and entirely your call: this PR is now 111 files and ~10.5k additions, and the date contract (`query_params.py`, `code_execution.py`, `create_data.py` prompt rules) is a distinct feature from browser verification that happens to have been found by it. It would be easier to review, revert and bisect on its own. Not blocking.

CI on this head: `unit-tests`, `playwright-tests`, `integration-llms`, `integration-data-sources` all green; both `e2e-tests` legs still running.

---

PR #1135, opened 2026-09-14
Add run_query tool to execute existing queries with parameter values

yochze:
## Summary

Introduces the `run_query` tool, enabling the AI agent to execute existing queries with caller-supplied parameter values. This complements `read_query` (which inspects query structure) by allowing the agent to answer "what about different inputs?" questions without creating new queries or modifying shared dashboards.

### Backend Implementation
- **New tool**: `RunQueryTool` in `app/ai/tools/implementations/run_query.py` (498 lines)
  - Executes saved query code with supplied parameter values in VIEWER mode
  - Returns cached results per (step, viewer, values fingerprint)
  - Never creates new Steps or modifies the query's default snapshot
  - Validates parameters: detects missing required values, unknown names, and identity-locked bindings
  - Provides structured `missing_params` output for the planner to request values from users

- **Input/Output schemas**: `app/ai/tools/schemas/run_query.py`
  - `RunQueryInput`: accepts `query_id` or `visualization_id`, parameter values, and optional force refresh
  - `RunQueryOutput`: returns success status, applied parameters, data preview, and missing parameter declarations
  - `MissingParam`: structured representation of required parameters needing values

- **Comprehensive E2E tests**: `tests/e2e/test_run_query_params.py` (379 lines)
  - Validates that saved code executes with requested values (not stored snapshot)
  - Confirms no new Steps are created and shared dashboards remain untouched
  - Tests per-viewer caching and repeat calls served from cache
  - Verifies missing required parameters trigger user prompts instead of execution
  - Ensures unknown parameters, undeclared params, and identity-locked values are rejected
  - Confirms execution failures never fall back to stored snapshot

- **Unit tests**: `tests/unit/test_queries_section_parameters.py`
  - Validates parameter declarations are rendered in query context sections

### Context & Observability
- Updated `persisted_summary.py` to include `run_query` in tool context summaries with applied parameters and cached status
- Enhanced `message_context_builder.py` to digest `run_query` results showing query ID and parameter values
- Modified `queries_section.py` to advertise declared parameters on queries, allowing the planner to see which queries accept values without spending a `read_query` call
- Updated `query_context_builder.py` to extract and format parameter specifications

### Frontend
- **New component**: `RunQueryTool.vue` (204 lines)
  - Displays execution status with parameter values in header
  - Shows data preview from THIS run (never re-hydrated from stored step)
  - Renders missing parameters and errors
  - Collapsible by default with cached indicator
  - Synthetic step prevents hydration from stored snapshot

- **Integration**: Registered in chat and report pages, added to groupable tools and tool labels

### Planner & Agent Updates
- Updated `prompt_builder_v3.py` to document `run_query` routing and usage
- Added `run_query` to `agent_v2.py` viewer-mode tools list
- Updated `review_producers.py` to include `run_query` in query tools tracking
- Added i18n strings for running/ran status labels with parameter display

- **Viewer mode execution**: Uses `QueryService.run_query_viewer()` to execute with resolved values while preserving the query's default step and shared dashboard state
- **Parameter validation**: Pre-checks for missing required parameters before execution, returning structured declarations so the planner can ask users rather than guessing
- **Identity-sourced parameters**: Explicitly excluded from caller input (e.g., `identity`, `input_identity_default`) and resolved server-side
- **Caching strategy**: Results cached per (step, viewer, values fingerprint) with optional force refresh
- **Error handling**: Distinguishes between parameter errors, execution failures, and missing context; never serves stored snapshot on failure
- **Report scoping**: Query IDs are addressable only from the report that owns them, matching `read_query` behavior
- **Dual ID support**: Accepts either `query_id` or `visualization_id` (resolves to the

https://claude.ai/code/session_01LqWeDYfH6iSi4yHWSU2bb9


yochze:
Reviewed commit `424c6f9c`. The `run_query` / `read_query` split is sound, but I recommend fixing these before merging:

1. **[P1] Expanding a result after reload can display the wrong parameter slice.** In `frontend/components/tools/RunQueryTool.vue:164–178`, the synthetic step ID does not prevent hydration: `ToolWidgetPreview` derives `queryId` from `result_json.query_id` and fetches `/api/queries/{qid}/default_step` whenever data is truncated. Report serialization caps results at 20 rows. For a report owner, running `region=DE` with more than 20 rows, reloading, and expanding the card can therefore replace the DE result with the saved US snapshot while the header still says DE. Downloads can follow the same hydration path. Add explicit run-result hydration that retrieves the matching execution’s data and preserves its applied parameters; cover reload, expansion, and export with a result larger than 20 rows.

2. **[P2] New labels are missing from Spanish and Hebrew.** `locales/en.json:1022` adds eight `tools.runQuery` strings, but neither `locales/es.json` nor `locales/he.json` contains them. These locales receive English fallback labels, and the catalogs no longer have the required matching shape. Add the corresponding translations.

Validation: exercised the PR’s actual serialization functions with a 25-row result and confirmed the 20-row, `truncated=true` output; traced the frontend hydration path; checked the English, Spanish, and Hebrew catalogs. Full-stack and browser tests were not run.

yochze:
Both findings confirmed. Fixing both.

**[P1] Real, and my earlier verification missed it.** I traced the same path you did:

- `persisted_summary.py:29` — `UI_TOOL_PREVIEW_ROWS = 20`, so a report-serialized tool preview over 20 rows carries `truncated: true`.
- `ToolWidgetPreview.vue:862` — `queryId` falls back to `result_json.query_id`. My synthetic execution spreads the whole result into `result_json`, so it carries the real `query_id`. The synthetic step id never gets consulted.
- `needsFullStep()` (:922) returns true on `data.truncated`; `hydrateExpandedCard()` (:931) then calls `hydrateLatestStep()` (:892), which GETs `/api/queries/{qid}/default_step` and overwrites `stepOverride` with the saved snapshot — the query's default values, not the run's.
- `addToSpreadsheet` (:1474) takes the same path, so export has the same defect.

I had claimed this was closed. It wasn't — my live check returned 1 row, which never reaches the truncation branch, so I verified the small-result case and generalized from it. That was wrong.

The fix is the explicit run-result hydration you describe, and the card already has the right mechanism to reuse: `runCardParams()` (:604) POSTs `/api/queries/{id}/run` with `{mode: 'viewer', params}`, which returns the per-viewer cached slice for those exact values (`StepUserResult`, keyed by params fingerprint — so a repeat is a cache hit, not a re-execution). I'll route run_query's hydration through that with the run's own `applied_params` rather than the default-step GET, keyed off an explicit prop so no other tool's hydration behavior changes. Tests will cover reload, expand, and export with a result larger than 20 rows.

**[P2] Confirmed.** `tools.runQuery` is in `locales/en.json` only. Adding the translations.

---

yochze:
Both fixed in `d1eabf7`.

**[P1]** Reproduced first, then fixed, measured with the same instrument on a real 36-row Germany slice of a 180-row query (full page reload, then expand):

| | endpoint on expand | card label | rendered rows |
|---|---|---|---|
| before | `GET …/default_step` | `180 rows` | US 6, Germany 6, France 4, Japan 3, Brazil 3 |
| after | `POST …/run` | `36 rows` | Germany 24 (all of them) |

Exactly your scenario: all five regions under a header still reading `region = Germany`.

The fix is the explicit run-result hydration. A viewer-run card re-requests its own slice via `POST /queries/{id}/run` in viewer mode with the values currently on the card — a lookup, not a re-execution, since the server caches per (step, viewer, params fingerprint) — and it returns every row rather than the 20-row preview, so **export is fixed by the same path**.

Two things I found while fixing that your note didn't cover:

- **Five refresh paths, not one.** Besides `hydrateExpandedCard` and `addToSpreadsheet`, the editor's `query:default_step_changed` broadcast, the `tool_preview:rebind` handler and `handleEntitySaved` each fetched `default_step` independently. All five now share one rule.
- **Identity params had to be stripped.** Replaying a server-resolved identity value back would make `resolve_param_values` reject every refresh. `input_identity_default` is an ordinary input and stays.

The decision lives in `utils/viewerRunHydration.ts` so it is unit-testable and so the five paths can't drift; an explicit `viewerRun` prop gates it, leaving every other tool's hydration untouched. It also follows the values on screen rather than the agent's original call, so a refresh after someone edits the card's parameter bar stays consistent.

Tests: `tests/unit/viewerRunHydration.mjs` (endpoint choice, identity stripping, on-screen values winning, null-value slices) and `backend/tests/unit/test_run_query_large_result_projection.py` (25-row result → 20 rows + `truncated`, `applied_params` surviving the cap — the precondition that made this reachable).

**[P2]** Fixed for **all nine** non-English locales, not just `es`/`he` — every other `tools.*` block exists in all ten, so the shape rule applies across the board. Verified all ten now carry `tools.runQuery` with identical keys.

One correction to my earlier claim on this PR: I said the hydration trap was closed and cited a screenshot. That screenshot was a 1-row result, which never crosses the cap and so never triggers a refetch. I verified the small case and generalized it — your reading of the code was right and mine wasn't.

---

yochze:
Confirmed and fixed in `ce94cc4`. Reproduced first, then measured after, same instrument — reload, expand, "Open in side panel" on the 36-row Germany slice of a 180-row query:

| | panel endpoint | rendered rows |
|---|---|---|
| before | `GET …/default_step` | US 9, Germany 9, France 7, Japan 6, Brazil 6 |
| after | `POST …/run` | Germany 36 |

**I took a different route than the one you suggested.** Carrying the context through the event → `panelData` → panel props fixes that one door. But there are four places a preview is mounted, and two more of them pass the raw execution with no props to thread: the grouped-block preview (`pages/reports/[id]/index.vue:429`) and the shared-link page (`pages/c/[token]/index.vue:209`). A prop-threading fix would have left both broken, and any mount point added later would reintroduce it.

So the slice is now read off the execution instead: a `run_query` result already carries `query_id` and `applied_params`, and `tool_name` distinguishes a slice from a snapshot. `deriveViewerRun()` replaces the prop, which is gone. All four mount points are correct without plumbing, including the two your note didn't reach.

Tools whose default step *is* their result — `create_data`, `describe_entity`, `read_query` — derive null and keep fetching it, so their hydration is untouched; that's asserted explicitly.

You were also right that the earlier tests didn't exercise the handoff — they tested the decision function, which would have returned the right request *if given* the context, while the defect was that nothing supplied it. `tests/unit/viewerRunHydration.mjs` now covers derivation from a panel-shaped `panelData.toolExecution`, the non-slice tools, a still-running execution, a failed run, and a defaults-only run.

One correction to your reproduction: it doesn't need the reload. The panel mounts a fresh component with its own state, so it starts from the serialized payload whenever it's truncated — the reload just makes it deterministic.

---

PR #1136, opened 2026-09-14
fix: chat bubble hidden by Intercom on shared dashboards; Polish prompt hides long text

r0533130357-dev:
Two small UI fixes.

## 1. Shared dashboard chat bubble hidden by Intercom

Intercom is booted by the app layouts and survives client-side navigation into `/r/{id}`. The two in-app routes there are:

- the post-sign-in redirect from the chat bubble's "Sign in"
- the 403 bounce from `/reports/{id}`

Its launcher then sits in the same bottom-end corner as the artifact chat bubble and the "Made with Bag of words" badge, above both in z-order, so the bubble is partly or fully covered. When `/r/{id}` is opened directly or in a new tab, Intercom isn't booted and there's no overlap, which is why it only happened sometimes.

**Fix:**
- `/r/{id}` measures Intercom's launcher (`useIntercomLauncherClearance`) and stacks the badge and bubble above it.
- They go above rather than under because the chat panel opens upward from the bubble, and a launcher on top would cover the open panel.
- With no launcher on screen (direct open, or mobile/Excel where it's hidden) the layout is unchanged.

## 2. Polish prompt cuts off long instructions

The "Polish this element" prompt was a one-line input next to the Apply button, so a long instruction scrolled sideways out of view.

**Fix:**
- It is now a textarea that grows to about 6 lines and then scrolls.
- Enter applies and Shift+Enter adds a line break.
- `dir="auto"` so Hebrew and other RTL text aligns correctly.
- The box is now placed below or above the picked element, kept inside the dashboard area. Before, it was clamped to fixed 500px/400px bounds and could be cut off.

## Testing

- `node frontend/tests/unit/intercomLauncher.mjs` passes (launcher clearance math).
- Both components compile.
- **Not verified in a browser.** Intercom only boots in production, so the overlap needs a check on staging: sign in via the chat bubble's "Sign in" and confirm the bubble sits above Intercom's launcher.
- The Polish prompt also still needs a manual check on a real dashboard, including an element near the bottom edge.
- The launcher is found by Intercom's own class/name hooks. If those are wrong or change, nothing matches and the page keeps today's layout.


frontend/components/dashboard/ArtifactFrame.vue
```diff
@@ -491,7 +496,7 @@
           </div>
 
           <!-- Modal Content - Full artifact iframe or SlideViewer -->
-          <div class="flex-1 min-h-0 relative bg-white dark:bg-gray-900">
+          <div ref="polishContainerRef" class="flex-1 min-h-0 relative bg-white dark:bg-gray-900">
```
yochze: `polishContainerRef` is on the wrong element, so the new clamping never runs — and the "stays inside the dashboard area" part of the fix doesn't take effect.

This div is the **fullscreen modal's** content pane: it lives inside `<Teleport to="body">` and is only mounted while `isFullscreenOpen` is true. The Polish box (line 445) is `absolute`-positioned inside the *inline* "Iframe Container" at line 219, and polish mode is inline-only — `enterPolishMode()` posts `POLISH_ENTER` to `iframeRef` (line 416, the inline iframe), while the fullscreen iframe at line 519 has no ref and never gets the message.

So whenever the box is visible, `polishContainerRef.value` is `null` and both fallbacks kick in:

```js
const containerWidth  = polishContainerRef.value?.clientWidth  ?? Infinity;
const containerHeight = polishContainerRef.value?.clientHeight ?? Infinity;
```

which collapses the math to `top = r.top + r.height + 8` and `left = r.left` — unclamped. The flip-above branch (`top + boxHeight > containerHeight - gap`) is also dead, since `containerHeight` is `Infinity`.

That's a regression for the exact case the PR targets: the old code clamped to `min(..., 500)` / `min(..., 400)`, which at least kept the box on screen for an element near the bottom; now it renders below the pane.

The fix is to move the ref to the container the box is actually positioned in (line 219), which is also the one the rect coordinates are relative to — the inline iframe is `absolute inset-0 w-full h-full` inside it, so `r.top`/`r.left` from `getBoundingClientRect()` in the iframe map 1:1 onto that div's box:

```diff
-    <div class="flex-1 min-h-0 relative bg-white dark:bg-gray-900">
+    <div ref="polishContainerRef" class="flex-1 min-h-0 relative bg-white dark:bg-gray-900">
```

and drop it from this modal div.

---

frontend/components/dashboard/ArtifactFrame.vue
```diff
@@ -457,19 +458,23 @@
...
+          <textarea
             ref="polishInputRef"
             v-model="polishInstruction"
-            type="text"
+            rows="2"
+            dir="auto"
             placeholder="e.g. make this bigger, change colors..."
-            class="flex-1 text-sm border border-gray-200 dark:border-gray-700 rounded-md px-2.5 py-1.5 focus:outline-none focus:ring-1 focus:ring-indigo-400 focus:border-indigo-400"
+            class="block w-full resize-none max-h-36 overflow-y-auto text-sm border border-gray-200 dark:border-gray-700 rounded-md px-2.5 py-1.5 focus:outline-none focus:ring-1 focus:ring-indigo-400 focus:border-indigo-400"
+            @keydown.enter.exact.prevent="submitPolishPrompt"
```
yochze: Enter-to-submit fires during IME composition, which truncates the instruction for CJK input.

A keypress that commits an IME candidate arrives as `keydown` with `keyCode 229` and no modifiers, so `.exact` matches it and `submitPolishPrompt()` runs on what is really a "confirm this candidate" keystroke. The user loses the rest of the sentence and the prompt is submitted half-written — and `submitPolishPrompt` calls `exitPolishMode()`, so there's no way back to the partial text.

This matters more than usual here since the change is explicitly about non-Latin input (`dir="auto"` for Hebrew); the same box is what a Japanese or Chinese user types into. The old `<input>` had the same hazard via implicit form submit, but a one-line input made it far less likely you'd be mid-composition on a long instruction.

`event.isComposing` is the guard:

```diff
-            @keydown.enter.exact.prevent="submitPolishPrompt"
+            @keydown.enter.exact="onPolishEnter"
```

```js
function onPolishEnter(e: KeyboardEvent) {
  if (e.isComposing) return;   // IME candidate commit, not a submit
  e.preventDefault();
  submitPolishPrompt();
}
```

---

frontend/composables/useIntercomLauncherClearance.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  function schedule() {
+    if (!frame) frame = requestAnimationFrame(measure)
+  }
+
+  onMounted(() => {
+    measure()
+    // Intercom injects (and on hide/show, restyles) its launcher asynchronously.
+    observer = new MutationObserver(schedule)
+    observer.observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['style', 'class'] })
```
yochze, line 30: Nit (not blocking): this observer is subscribed on every shared dashboard, but the condition it watches for is rare.

`subtree: true` + `attributes: true` over `document.body` means every `class`/`style` change anywhere on the page — chart re-renders, spinners, hover states, the top bar toggling — schedules a frame that runs `querySelectorAll` over the whole document. The `rAF` coalescing keeps it to one pass per frame, and with no launcher present `getBoundingClientRect` is never called, so it's bounded. But `/r/{id}` is the page most likely to be embedded and left open on a wall display, and Intercom is only ever there when the viewer arrived by client-side nav from the app in production — every other visit pays the observer for nothing.

Cheap narrowing, if you think it's worth it: start with `childList` only, and widen to attributes once a launcher node actually matches (the `hide_default_launcher` restyle only matters after the element exists). Happy for this to stay as is if you'd rather keep the composable simple.

---

yochze (changes requested):
Reviewed both fixes. The Intercom half is sound; the Polish half has one wiring bug that makes the repositioning a no-op (and a regression for bottom-edge elements), plus an IME issue on the new Enter-to-submit.

## 1. Intercom clearance — looks right

The approach (measure the DOM rather than assume Intercom's size) is the correct call given `hide_default_launcher` toggles on mobile/Excel and the launcher swaps shapes mid-load. Checks I ran:

- **Math.** `launcherClearance` returns distance from the viewport bottom to the launcher's top. Default launcher (48px, 20px inset, 800px viewport) → 68; `cornerLift` = 60; badge lands at `bottom: 80px`, 12px clear of the launcher top. Bubble at `16 + 60 = 76` when there's no badge, 8px clear. Both correct.
- **Degradation.** No launcher → `clearance` 0 → `cornerLift` 0 → byte-identical layout to today. Hidden launcher measures 0 tall and is skipped. Selector drift fails the same way. The fallback story in the PR description holds up.
- **Panel height.** `calc(100dvh - ${bottomPx + 96}px)` reproduces the old `7rem` exactly in the unraised case and fixes the raised case, which previously under-reserved (64px bottom against a 112px subtraction). Good catch as a drive-by.
- **SSR.** `onMounted`-only measurement with `clearance` starting at 0 means server and client first paint agree — no hydration mismatch.
- **RTL.** Intercom's `alignment` flips to left in RTL (`layouts/default.vue:1340`) and the page's `end-*` flips with it, so the two stay in the same corner and a vertical-only lift is the right shape for both.
- The unit test passes here (`node frontend/tests/unit/intercomLauncher.mjs`, Node 22.22) and its cases cover the ones that matter: empty, default, hidden, two-shapes-at-once, below-fold.

One gap worth noting: the half that got a test is the simple half. `polishPromptPosition` is now genuinely non-trivial geometry (flip above/below, two-axis clamping, a height that changes as you type) and has no coverage — extracting it the same way `launcherClearance` was extracted would have caught the bug below.

## 2. Polish prompt — the container ref is on the wrong element

`ref="polishContainerRef"` is on the fullscreen modal's pane (inside a `Teleport`, mounted only while the modal is open), but the Polish box is positioned inside the inline iframe container, and polish mode only ever runs inline. So the ref is `null` whenever the box is visible, both dimensions fall back to `Infinity`, and every clamp — including the flip-above-the-element branch — becomes a no-op. Net effect: `top = r.top + r.height + 8`, `left = r.left`, unbounded. Since the old code did clamp (to 500/400), an element near the bottom of a tall artifact is worse off than before. Details and the one-line fix inline.

The rest of that change is fine — the `scrollHeight + border` auto-grow is correct under `box-sizing: border-box`, `max-h-36` does work out to ~6 lines at `text-sm`, and re-picking an element resets the height because `polishInstruction` actually changes. Second inline comment covers Enter-vs-IME.

## Verification

Given the PR is explicit that neither fix was seen in a browser, when you do get to staging it's worth confirming the Polish box specifically on an element in the **bottom** ~150px of a dashboard, in both the pre- and post-fix states — that's the case the clamping exists for and the one the wrong ref silently disables.

---

PR #1139, opened 2026-09-14
feat(agents): browse and preview file connections in the Files tab

r0533130357-dev:
## Summary

The agent **Files** tab showed the first 30 names of each SharePoint / Google Drive / network share connection as a flat list. It is now a file browser, and clicking a file previews it in place.

**Browse**
- Folder tree with breadcrumbs, search across the whole scope (a hit shows its folder and jumps there), size, modified date, "open in source" for SharePoint/Drive.
- Shows exactly what the agent can see: same live listing as `list_files`, include-globs applied, off-scope folders never appear.
- Listing errors show with a retry instead of reading as "No files match".

**Preview**
| Type | Shown as |
|---|---|
| PDF | the shared `FilePreview` viewer (fit-width, no thumbnail rail) |
| Word / PowerPoint | converted to PDF with the LibreOffice converter `read_file` already uses |
| Excel / CSV | `FileTablePreview` — first 200 rows, real row count, sheet tabs |
| Images | `FilePreview`; animated APNG/GIF/WebP frozen on the first frame, as SharePoint's viewer does |
| Text (txt, md, json, …) | text |
| Other / > 25 MB | "no preview" with "open in source" (oversize files are never requested) |

Opening a file scrolls the pane so the whole preview is on screen.

## Connector support

All ten `data_shape: files` connectors mount the browser; what they get differs:

| Connector | Folder tree | Preview |
|---|---|---|
| SharePoint (incl. all libraries), OneDrive, Google Drive, network share, S3, SharePoint Server, Documentum | ✓ | ✓ |
| OneNote | ✓ (notebook / section / page) | ✓ — page HTML, sandboxed |
| Outlook mail, Gmail | flat list (subjects are never split into folders) | — no raw reader; "open in source" only. Gmail lists its 25 most recent messages. |

## Review follow-up (68ae3f6)

- Root/prefix escapes (network_dir, S3) are now a `ScopeEscapeError` (a `GlobScopeError`): 403 + `file.access_denied` in the preview, and audited by read_file / attach_file / write_file / grep_files too. The 403 no longer echoes the path.
- **Pre-existing bug fixed:** SharePoint on all libraries with include-globs denied in-scope files on any read from a fresh client (every request, `read_file` included) — `_library_name` read the drive map without resolving it. Test fails without the fix.
- Size is checked before download on every connector (`max_bytes` threaded through `read_source_bytes`; Graph, Drive, S3, network_dir now reject from the reported size). LibreOffice conversions capped at 2 per worker.
- Blob-URL leak on the image path fixed; animation sniffed from the first 64 KB. 403 shows the neutral box. Listing echoes the effective limit.
- Tests: the listing/preview contract runs per connector (SharePoint all-libraries with a fresh client per request, S3, Documentum, SharePoint Server), plus audit rows for both denial shapes, oversize-before-read, and a no-raw-reader connection.

## Backend

- `GET /data_sources/{id}/connections/{cid}/files` now returns `path` and `web_url`, and up to 5000 files per call (was 500). Connectors list files only, so the tree is built client-side from `path`; every call already walked the whole scope, so paging just repeated the walk.
- New `GET /data_sources/{id}/connections/{cid}/files/content?file_id=…[&format=table|pdf][&sheet=…]`:
  - Reads live through the connection client and `read_source_bytes` — the same path and glob check as the agent's `read_file`. Off-scope → 403 + `file.access_denied` audit; each view → `file.viewed` audit. Per-user connections read with the viewer's credentials. Nothing is copied into BOW.
  - 25 MB cap (`_PREVIEW_MAX_BYTES`, shared with `read_file`). Table parsing and Office conversion run off the event loop.
  - Raw bytes are sent as an attachment with `X-Content-Type-Options: nosniff` and `Content-Security-Policy: sandbox`. The browser builds its own typed blob from the extension, so an iframe only ever gets a PDF and an `` only an image.
  - `format=pdf` returns 422 when `soffice` is absent (it is in the Docker image; not on a bare dev machine).

## Frontend

- `ConnectionFileBrowser.vue` (browser) and `ConnectionFilePreview.vue` (fetch + states), mounted in `AgentFilesPanel`.
- `FilePreview.vue` gains an optional `src` prop (content already loaded) so the browser reuses `read_file`'s viewer. Without `src` its behaviour is unchanged.
- `utils/fileTree.ts`: tree building, search, preview kind by extension, animated-image detection (header-only). `fileIcon` moved here from `KnowledgeExplorer`, unchanged.
- 17 new strings in all 10 locales.

## Testing

- `backend/tests/e2e/test_connection_file_browse.py` (8): scope-relative paths and off-scope hiding, >500 files in one call, raw preview headers, 403 off-scope and `../` escape, CSV row cap vs real count, Excel blank-row trimming and sheet switching, non-tabular rejection, Office→PDF (converter mocked) and the 422 path. The listing tests fail on the old endpoint.
- Related suites pass: Documentum, SharePoint on-prem, file-agent tables view, read_file source document — 54 passed.
- `frontend/tests/unit/fileTree.mjs`: tree, ordering, odd paths, search, preview kinds, APNG/GIF/WebP detection.
- Driven end to end with Playwright against a sandbox (network_dir with nested folders): navigation, breadcrumbs, search → open → close back in the hit's folder, every preview type, a real Word file converted with LibreOffice, a real 31-frame APNG from SharePoint shown frozen, oversize file never fetched, auto-scroll keeps the preview fully visible at 900px, Hebrew/RTL, audit rows written.

**Not verified:** SharePoint (Graph) through the UI end to end against a live tenant — its listing, raw read and scope check run through the real client against a simulated Graph in the tests, and the APNG came from a real tenant, but the UI runs were on network_dir. The `read_file` chat card after the `FilePreview` change was not re-run with a live LLM (the no-`src` code path is unchanged).


frontend/components/datasources/ConnectionFilePreview.vue
```diff
@@ -0,0 +1,176 @@
...
+    }
+    // Type the blob ourselves from the extension: an iframe only ever gets a
+    // PDF and an <img> only an image, whatever the source claimed.
+    const type = k === 'image' ? (IMAGE_MIME_BY_EXT[fileExt(name.value)] || 'application/octet-stream') : 'application/pdf'
+    let typed = new Blob([blob], { type })
+    if (k === 'image' && isAnimatedImage(new Uint8Array(await blob.arrayBuffer()))) {
+      typed = await firstFrame(typed)
+      if (mine !== seq) return
+    }
+    url.value = URL.createObjectURL(typed)
```
yochze: Two things on these five lines.

**1. Blob URL leaked on unmount for _non-animated_ images.** `onBeforeUnmount` bumps `seq` specifically so that "a response still in flight is dropped instead of minting a blob URL nobody revokes", and every other path re-checks after its await — but this one doesn't. When `isAnimatedImage` returns `true` there's a guard after `firstFrame`; when it returns `false` we fall straight from `await blob.arrayBuffer()` to `URL.createObjectURL(typed)` with no re-check. Close the preview while an image is still decoding and the object URL is minted on a destroyed instance, so `clear()` has already run and nothing ever revokes it — up to 25 MB pinned for the lifetime of the page.

**2. The whole file is buffered just to sniff a container header.** `isAnimatedImage` never reads past the header (≤4 KB for the GIF app-extension scan, the pre-`IDAT` chunks for PNG, 21 bytes for WebP), but `await blob.arrayBuffer()` materializes a second full copy of a 25 MB image to feed it.

Both fall out of one edit:

```js
    let typed = new Blob([blob], { type })
    if (k === 'image') {
      // Header-only: isAnimatedImage never reads past the container header.
      const head = new Uint8Array(await blob.slice(0, 65536).arrayBuffer())
      if (mine !== seq) return
      if (isAnimatedImage(head)) {
        typed = await firstFrame(typed)
        if (mine !== seq) return
      }
    }
    url.value = URL.createObjectURL(typed)
```

64 KB is generous headroom for a PNG carrying an `iCCP` profile or `eXIf` block ahead of `acTL`.

---

backend/app/routes/data_source.py
```diff
@@ -624,18 +628,152 @@ async def list_connection_files(
...
+        content, name, mime = await read_source_bytes(client, file_id)
+    except GlobScopeError as e:
+        await audit_file_access_denied(
+            {"db": db, "organization": organization, "user": current_user}, connection_id, file_id, str(e),
+        )
+        raise HTTPException(status_code=403, detail=str(e))
+    except HTTPException:
+        raise
+    except Exception as e:
+        raise HTTPException(status_code=400, detail=f"Failed to read file: {e}")
```
yochze, line 752: **A path-escape attempt is the one denial that produces no audit row.**

The clients split their denials across two exception types at the same chokepoint:

- `network_dir_client._resolve` — off-glob raises `GlobScopeError`, but escaping the root raises a plain `ValueError("Path escapes the connection root: …")`.
- `s3_client._resolve_key` — off-glob raises `GlobScopeError`, but escaping the prefix raises a plain `ValueError("Key escapes the connection prefix: …")`.

Only `GlobScopeError` lands here, so only the *in-scope-but-off-glob* case gets 403 + `file.access_denied`. The genuinely hostile input — `../../etc/passwd`, a key climbing out of the bucket prefix — falls through to the generic handler below and returns `400 Failed to read file: Path escapes the connection root: ../../etc/passwd`: no audit row, and the attacker's path echoed back in the body. That inverts the priority you'd want in the trail.

The test acknowledges it rather than pinning it:

```python
r = test_client.get(content, params={"file_id": f"../outside-{tmp_path.name}.txt"}, headers=headers)
assert r.status_code in (400, 403), r.text
```

Suggest catching both and collapsing the detail:

```python
    except (GlobScopeError, ValueError) as e:
        await audit_file_access_denied(
            {"db": db, "organization": organization, "user": current_user}, connection_id, file_id, str(e),
        )
        raise HTTPException(status_code=403, detail="This file is outside the connection's scope.")
```

Then the test can assert `403` flatly, and both denial shapes show up under one action in the audit trail. Worth checking the same split in `read_file`/`attach_file` while you're here — they route through the identical `audit_file_access_denied` and look to have inherited it.

---

backend/app/routes/data_source.py
```diff
@@ -624,18 +628,152 @@ async def list_connection_files(
...
+
+    if len(content) > _PREVIEW_MAX_BYTES:
+        raise HTTPException(status_code=413, detail="File is too large to preview.")
```
yochze, line 757: **The cap is enforced after the download, and only the frontend guards the request.**

By the time this runs the bytes are already fully resident in the worker. The frontend does check `size` before asking (`MAX_PREVIEW_BYTES` in `ConnectionFilePreview`), which covers normal use, but the endpoint takes an arbitrary `file_id` over plain HTTP and the connectors don't all backstop it:

- `graph_drive_client.read_raw_bytes` (SharePoint / OneDrive) has no byte cap at all.
- `network_dir_client.read_raw_bytes` checks `self.max_file_bytes`, which is `None` unless `max_file_mb` is configured on the connection.
- `google_drive_client.read_raw_bytes` likewise takes no limit.

So one GET for a 2 GB in-scope SharePoint file pulls the whole thing into memory before we reject it. `read_file` shares the shape, but there it's LLM-paced behind a chat turn, not a hand-crafted GET any org member with `data_source:view` can issue.

Cheapest fix that doesn't touch the clients: the listing already returns `size`, so resolve the entry first and reject before reading when the connector reports one. Better still, thread `_PREVIEW_MAX_BYTES` down — `documentum_client.read_raw_bytes` and `sharepoint_onprem_client.read_raw_bytes` both already accept `max_bytes=`, so `read_source_bytes` could forward it where the signature allows.

---

backend/app/routes/data_source.py
```diff
@@ -624,18 +628,152 @@ async def list_connection_files(
...
+            # pandas parsing is CPU-bound — a 20 MB workbook would stall this worker's event loop.
+            return await asyncio.to_thread(_table_preview, content, name, sheet)
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"Could not read the table: {e}")
+
+    if format == "pdf":
+        from app.data_sources.clients._office_convert import CONVERTIBLE_EXTS, office_to_pdf_bytes
+        if name.rsplit(".", 1)[-1].lower() not in CONVERTIBLE_EXTS:
+            raise HTTPException(status_code=400, detail="PDF preview is only available for Office documents.")
+        converted = await asyncio.to_thread(office_to_pdf_bytes, content, name)
```
yochze: **Unbounded concurrency on `soffice`.** Each call spawns a LibreOffice subprocess with a 90 s timeout (`DEFAULT_TIMEOUT_SECONDS`) and its own profile directory. The only ceiling is the default `asyncio.to_thread` executor, ~`min(32, cpu+4)`, so a handful of users clicking through a folder of `.docx` files can hold 32 concurrent LibreOffice processes for up to 90 s each.

This path existed before for `read_file`'s vision fallback, but that's agent-paced — one conversion per tool call, inside a turn. The browser makes it click-paced: every click on an Office file in the tree is a conversion, and nothing dedupes a double-click on the same file. A module-level `asyncio.Semaphore` around the `to_thread` call (or a short-TTL cache keyed on connection + file id + version) would bound it without changing the UX in the common case.

---

frontend/components/datasources/AgentFilesPanel.vue
```diff
@@ -44,24 +44,8 @@
...
-        </div>
-        <div v-else class="text-xs text-gray-500 dark:text-gray-400 mb-1">{{ browse[conn.id]?.total ?? 0 }} files match · agent reads ONLY these · denials audited</div>
-        <ul class="text-xs font-mono text-gray-600 dark:text-gray-400 space-y-0.5 max-h-48 overflow-auto">
-          <li v-for="n in (browse[conn.id]?.names || [])" :key="n" class="truncate">{{ n }}</li>
-          <li v-if="(browse[conn.id]?.total || 0) > (browse[conn.id]?.names?.length || 0)" class="text-gray-400 italic">… {{ browse[conn.id].total - browse[conn.id].names.length }} more</li>
-          <li v-if="browse[conn.id] && browse[conn.id].total === 0" class="text-gray-400 italic">{{ indexModeOf(conn) === 'none' ? 'Live source — read on demand, not cached.' : 'No files match.' }}</li>
-        </ul>
-        </template>
+      <div class="mt-3 border-t border-gray-100 dark:border-gray-800 pt-3">
+        <ConnectionFileBrowser :ds-id="dsId" :connection-id="conn.id" />
```
yochze, line 48: **The browser mounts for every `data_shape === 'files'` connector, and four of the ten aren't file stores.**

`fileConnections` filters on `registryByType[c.type]?.data_shape === 'files'`, which in `data_source_registry.py` is ten types:

`network_dir`, `s3`, `documentum`, `sharepoint_onprem`, `sharepoint`, `onedrive`, `google_drive`, `onenote`, `outlook_mail`, `gmail_mail`

The old flat name list degraded gracefully across all ten — a list of subjects or page titles still reads as a list. A folder tree keyed on `path` does not.

**Mail (`gmail_mail`, `outlook_mail`) — the tree is built from subject lines.** Both clients set `path` to the subject:

```python
# gmail_mail_client._message_to_item
"name": subject,
"path": subject,
```
```python
# graph_mail_client._msg_to_item
"name": m.get("subject") or "(no subject)",
"path": m.get("subject") or "(no subject)",
```

`pathSegments` splits on `/`, so an email subject like `Re: Q3 / Q4 forecast` becomes a folder `Re: Q3 ` containing a file ` Q4 forecast`. Subjects with slashes are common (`FW: Contracts/2025`, `A/B test results`), so mail connections render a tree of phantom folders. Two further mismatches: `gmail_mail_client._list_messages` is hard-capped at `maxResults: 25`, so "N files in scope" is always ≤25 regardless of mailbox size; and neither client implements `read_raw_bytes`, so `read_source_bytes` falls through to `aread_file` and serializes the message — while `previewKind` is computed from the *subject*, so a message titled `Q3 report.pdf` is typed `application/pdf` and handed to the PDF iframe as serialized message text.

**OneNote — tree is fine, preview never is.** `graph_onenote_client` builds a real notebook/section/page `path` and enforces globs on it, so the tree is meaningful. But page titles carry no extension, so `previewKind` is always `'none'` and every page shows "Preview isn't available for this file type" — even though `read_raw_bytes` returns perfectly renderable HTML and the entry already declares `"mime_type": "text/html"`.

Suggest gating on capability rather than `data_shape`: either narrow `fileConnections` to the six real file stores and keep the old flat list for mail/OneNote, or drive `previewKind` off `mime_type` when the extension yields nothing (which fixes OneNote for free) and skip tree-building when `path === name` for every entry (which collapses mail back to a flat list).

Worth being explicit in the PR description about which of the ten are actually supported — right now it reads as if all file connections got the browser.

---

backend/tests/e2e/test_connection_file_browse.py
```diff
@@ -0,0 +1,189 @@
...
+    user = create_user()
+    token = login_user(user["email"], user["password"])
+    org_id = whoami(token)["organizations"][0]["id"]
+    headers = {"Authorization": f"Bearer {token}", "X-Organization-Id": org_id}
+    ds = create_data_source(
+        name="Shared drive", type="network_dir",
+        config={"root_path": str(tmp_path), **config},
+        credentials={"auth_type": "none"}, user_token=token, org_id=org_id,
+    )
+    conns = test_client.get(f"/api/data_sources/{ds['id']}/connections", headers=headers).json()
```
yochze, line 23: **All eight tests run on `network_dir`** — every one goes through `_agent_on_dir`, which hardcodes `type="network_dir"`. That's the easiest connector and the least representative: its ids *are* paths, its `path` is already scope-relative, and its scope check is a filesystem `resolve()`.

The contract these tests pin ("`path` is returned and scope-relative, off-scope files never appear") is exactly the contract that varies most between connectors, because each one derives `path` differently:

- `s3_client._entry` — `path` is the key relative to `self.prefix`, and `_resolve_key` denies off-prefix with `ValueError` but off-glob with `GlobScopeError` (see my comment on the route's `except`).
- `graph_drive_client` — ids are opaque (`01TP3T7WAPS6…`) and multi-library connections hand out qualified `drive|item` ids; `path` comes from `_scoped_path(drive_id, parentReference.path, name)` and is *library-prefixed* when the connection spans every library. That prefixing is the single thing the tree's root level depends on, and it's untested.
- `sharepoint_onprem_client._scope(path, libraries)` — same idea, different derivation.
- `documentum_client` — `path` is a display path, `id` is `r_object_id`; the two are unrelated, which is precisely the case `pathSegments`' id-fallback was written for.

There are ready harnesses for at least three of these, so the cost is low:

- `tests/unit/test_sharepoint_onprem_client.py` exports the `remote` fixture (simulated SharePoint HTTP server) — `tests/e2e/test_sharepoint_onprem.py` already reuses it for real API/DB flows.
- `tests/unit/test_s3_client.py` has an S3 stub.
- `tests/e2e/test_documentum.py` exists.

The two listing assertions are the ones worth parameterizing across connectors — `path` present and scope-relative, and off-scope never listed. `test_browse_returns_more_than_500_files_in_one_call` and the preview-format tests can stay on `network_dir`; they're testing the route, not the connector.

Separately, `test_preview_denies_files_outside_the_scope` asserting `status_code in (400, 403)` should tighten to `403` once the escape/off-glob split is unified.

---

frontend/components/datasources/ConnectionFilePreview.vue
```diff
@@ -0,0 +1,176 @@
...
+  `/data_sources/${props.dsId}/connections/${props.connectionId}/files/content?` +
+  new URLSearchParams({ file_id: props.file.id, ...params }).toString()
+
+// A blob request's error body arrives as a Blob too.
+async function showError(err: any) {
+  let d = err?.data
+  if (d instanceof Blob) { try { d = JSON.parse(await d.text()) } catch { d = null } }
+  const detail = typeof d?.detail === 'string' ? d.detail : ''
+  const status = err?.statusCode ?? err?.status
+  if (status === 422 || status === 413) serverUnavailable.value = detail || t('agentsPage.fileBrowserPreviewUnsupported')
```
yochze: Nit: 403 falls through to `error.value`, which renders the red box with a **Try again** button. An off-scope denial can never succeed on retry, so it belongs with the 422/413 neutral-box cases.

Unreachable today — off-scope files aren't listed, so you can't click one — but it becomes reachable the moment a connection's include-globs are narrowed while someone has the browser open, which is also exactly when you'd least want a retry button inviting repeated `file.access_denied` rows into the audit trail.

```js
  if (status === 422 || status === 413 || status === 403) serverUnavailable.value = detail || t('agentsPage.fileBrowserPreviewUnsupported')
```

---

backend/app/routes/data_source.py
```diff
@@ -624,18 +628,152 @@ async def list_connection_files(
...
     return {
         "connection_id": str(connection_id),
-        "files": [{"id": f.get("id"), "name": f.get("name"), "size": f.get("size"),
-                   "modified_at": f.get("modified_at"), "mime_type": f.get("mime_type")} for f in page],
+        "files": [{"id": f.get("id"), "name": f.get("name"), "path": f.get("path"),
+                   "size": f.get("size"), "modified_at": f.get("modified_at"),
+                   "mime_type": f.get("mime_type"), "web_url": f.get("web_url")} for f in page],
         "total": total,
         "offset": offset,
         "limit": limit,
```
yochze: Nit: the slice clamps to `min(limit, 5000)` but the response echoes the caller's raw `limit`, so a client asking for `limit=100000` is told its limit was honoured while getting 5000 rows. `has_more` is computed from `len(page)` so it stays correct — it's only the echoed field that lies. Echoing the effective value would make the truncation self-describing:

```python
    effective = max(1, min(limit, 5000))
    page = files[max(0, offset): max(0, offset) + effective]
    ...
        "limit": effective,
```

Also worth noting for the frontend: `ConnectionFileBrowser` requests exactly `LIMIT = 5000` and then renders `fileBrowserTruncated` when `total > files.length`, telling the user to "narrow the connection's scope to browse the rest". Since 5000 matches `max_catalog_objects`, that banner should be rare — but a connection configured above that default will hit it with no way to page past it, because the browser never sends `offset`. Fine as a deliberate trade-off, just make sure the string is the intended dead end.

---

yochze (commented):
Reviewed the full diff (19 files, +1236/−57) against `main` at `45a6ebd`.

Solid PR, and the security posture on the new content endpoint is genuinely careful. Reusing `read_source_bytes` + `GlobScopeError` means the preview inherits `read_file`'s exact scope check instead of inventing a second one; `attachment` + `nosniff` + `CSP: sandbox` on the response, plus retyping the blob client-side from the extension, is the right belt-and-braces for untrusted source bytes; and the CPU-bound pandas/LibreOffice work is correctly pushed off the event loop. Deciding `previewKind` by extension rather than provider mime is the right call and the reasoning in the comment is worth keeping.

**What I verified**

- Every cross-module symbol the new code reaches for exists and matches: `_trim_to_data`, `office_to_pdf_bytes`/`CONVERTIBLE_EXTS`, `_content_disposition`, `_PREVIEW_MAX_BYTES`, `audit_service` at module scope, `xlrd`/`openpyxl` in `pyproject.toml`.
- The three ext→kind maps agree across the stack: frontend `table` == backend `_PREVIEW_TABLE_EXTS`, frontend `office` == `CONVERTIBLE_EXTS`, and every `image` ext has an entry in `IMAGE_MIME_BY_EXT`.
- All six real file clients emit `path` and `web_url`, so the tree does work for Graph/Drive/S3/Documentum/SharePoint-on-prem — the part the description flags as UI-unverified. `isWebUrl`'s `^https?://` filter correctly hides `file://` (network_dir) and `s3://` (S3) links.
- The 17 new locale keys exist in all 10 locales with matching interpolation placeholders; `countFiles` already had plural forms.
- `node frontend/tests/unit/fileTree.mjs` passes.
- `has_more` arithmetic is correct; the `office_to_pdf_bytes` monkeypatch in the new test works because the route imports it function-locally.

I could **not** run `backend/tests/e2e/test_connection_file_browse.py` — pytest isn't installed in my environment, so those 8 tests are unverified here.

**Main themes in the inline comments**

1. **Connector coverage** (`AgentFilesPanel.vue`) — the browser mounts for all ten `data_shape === 'files'` types, but `gmail_mail`/`outlook_mail` set `path` to the email *subject*, so any subject containing `/` produces phantom folders, and OneNote pages can never preview despite returning renderable HTML. The old flat list degraded gracefully here; a path-keyed tree doesn't.
2. **Test coverage** (`test_connection_file_browse.py`) — all eight tests run on `network_dir`, the connector whose ids already *are* paths. The listing contract being pinned is exactly the one that varies per connector; there are existing harnesses for SharePoint-on-prem, S3 and Documentum to reuse.
3. **Audit gap** (`data_source.py`) — path-escape raises plain `ValueError` in both `network_dir` and `s3`, so only the off-glob case gets 403 + `file.access_denied`. The more hostile input gets a 400 with the attacker's path echoed back and no audit row. The test's `assert status_code in (400, 403)` is the tell.
4. **Blob URL leak** (`ConnectionFilePreview.vue`) — the non-animated image path is the one branch missing its post-await `seq` re-check, which is the exact thing `onBeforeUnmount`'s `seq++` was written to prevent.
5. Resource bounds — the 25 MB cap lands after the download and several connectors have no cap of their own; `format=pdf` has no concurrency bound on `soffice`.

Plus a few nits. Nothing here blocks merge; #1 and #2 are the ones I'd want settled before this ships, since they're about what the feature claims to support.

---

backend/app/routes/data_source.py
```diff
@@ -624,18 +630,172 @@ async def list_connection_files(
...
+    """
+    from app.services.connection_service import ConnectionService
+    from app.data_sources.clients._file_source_common import FileTooLargeError, GlobScopeError
+    from app.ai.tools.implementations._file_tool_common import (
+        _PREVIEW_MAX_BYTES, audit_file_access_denied, read_source_bytes,
+    )
+
+    conn = await _attached_connection(db, data_source_id, connection_id, organization)
+    client = await ConnectionService().construct_client(db, conn, current_user)
+    if not hasattr(client, "read_raw_bytes"):
```
yochze, line 733: Hoisting `construct_client` out of the `try` to run the `read_raw_bytes` check is right, but it lost the error handling it used to sit inside. Before 68ae3f6 this line was the first statement of the `try`, so anything it raised was caught by `except Exception → 400`. Now a non-`HTTPException` from client construction escapes as an unhandled 500.

`ConnectionService.construct_client` has several such paths: `json.loads(connection.config)` on a malformed config, `resolve_client_class` on an unknown type, `ClientClass(**params)` raising `TypeError` when saved config no longer matches the client's signature. The per-user-OAuth 403 still propagates fine because it *is* an `HTTPException` — it's the rest that changed shape.

The sibling listing endpoint still wraps its own `construct_client` call, so the two endpoints now disagree on how a broken connection surfaces: 400 from `…/files`, 500 from `…/files/content`.

```python
    try:
        client = await ConnectionService().construct_client(db, conn, current_user)
    except HTTPException:
        raise
    except Exception as e:
        raise HTTPException(status_code=400, detail=f"Failed to open the connection: {e}")
    if not hasattr(client, "read_raw_bytes"):
        ...
```

---

frontend/components/datasources/ConnectionFilePreview.vue
```diff
@@ -0,0 +1,200 @@
...
+    if (mine !== seq) return
+    if (res.error.value) { await showError(res.error.value); return }
+    const blob = res.data.value as Blob
+    if (k === 'text' || k === 'html') {
+      const s = await blob.text()
+      if (mine !== seq) return
+      if (k === 'html') {
+        html.value = HTML_CSP + s.slice(0, MAX_HTML_CHARS)
+        return
+      }
```
yochze, line 171: Nit: the HTML path truncates silently, where the text path right below it sets `textTruncated` and renders `fileBrowserPreviewTextTruncated`. A OneNote page past 2 M chars just stops mid-document with nothing saying so — and unlike the text view, a truncated HTML render doesn't *look* truncated, since the parser closes the open tags for you.

Reusing the existing flag and string would cover it:

```js
      if (k === 'html') {
        html.value = HTML_CSP + s.slice(0, MAX_HTML_CHARS)
        textTruncated.value = s.length > MAX_HTML_CHARS
        return
      }
```

though the string interpolates `MAX_TEXT_CHARS`, so it'd need the applicable cap passed in.

---

yochze (commented):
Reviewed 68ae3f6 + 4679627. All eight threads from the last round are addressed — I've resolved them. Two new small things inline, neither blocking.

**The fixes hold up.** Checked rather than assumed:

- `ScopeEscapeError(GlobScopeError)` is the right shape — subclassing means every existing `except GlobScopeError` picks up escapes for free, and `FileTooLargeError(ValueError)` keeps the agent tools' existing `except ValueError` behaviour unchanged. The `test_grep_files` update from `not_found` → `access_denied` is a real improvement that fell out of it.
- `read_source_bytes`' `inspect.signature` probe is keyword-only (`*, max_bytes`) on every reader it touches, and no existing caller passes positionally, so nothing broke. The readers without the parameter fall back to the post-read backstop, which the comment says.
- `_enforce_item_scope`'s early return is semantically identical to the old inline call — `_enforce_scope` already no-ops without globs — so 4679627 removes a `/drives` round-trip and nothing else.
- `_OFFICE_PREVIEW_SLOTS` at module scope is fine on 3.12: `Semaphore` binds a loop only on first *contention*, and production is one loop per worker.

**The `_library_name` bug is the find of this round.** An all-libraries SharePoint connection with include-globs denying in-scope files on every read, `read_file` included, because `_drives` was `None` on a fresh client — that's a live access bug affecting the agent, not just the browser, and it's the kind of thing that only surfaces when a test builds a new client per request the way `_serve` does. Good catch, and the right test to have written.

**On the HTML preview** (new surface, so I looked at it fresh): `sandbox=""` plus `default-src 'none'` is the right pairing, and belt-and-braces — scripts are blocked twice over. I did wonder whether prepending the CSP meta ahead of OneNote's `<!DOCTYPE html>` would drop the document into quirks mode, so I checked it in Chromium rather than guess: it doesn't. `srcdoc` documents inherit the parent's mode, `compatMode` stays `CSS1Compat`, and the meta still lands as `head`'s first element child, so the policy applies. No change needed.

**Connector coverage** is now real: four connectors exercised through their own clients against simulated remotes, with `_serve` forcing a fresh client per request. `stub.assert_no_pending_responses()` after the S3 oversize case is a nice way to prove the `get_object` never happened. Gmail's 25-message cap stays, but it's documented in the description now, which is a fair resolution.

**CI:** `integration-data-sources`, `integration-llms` and `playwright-tests` are green; `unit-tests` and both `e2e-tests` legs were still running when I looked. The e2e legs are the ones carrying the new per-connector tests, so worth a glance before merge — I haven't been able to run pytest in my environment either round.

---

PR #1149, opened 2026-09-16
fix(mcp): request the scopes the MCP server demands, not the identity provider's

yochze:
Follow-up to #1140. With the host allowlist gone, the first DCR connection to an MCP server behind a **generic identity provider** (Auth0 / Okta / Entra) got past `oauth/authorize`, through consent, and then failed at the post-exchange connection test with `Failed to connect to MCP server: Client error '401 Unauthorized'`.

## Root cause

`discover_mcp_oauth` took `scopes_supported` from the **authorization-server** metadata. Behind a generic IdP that list is the standard OIDC set (`openid profile email offline_access …`) and says nothing about the resource. The token came back without the server's scope (`mcp:read`) and the MCP server rejected it. The catalog presets never showed this: their MCP server and AS are the same vendor, so both documents agree.

## Fix

**Discovery** (`mcp_dcr_service.py`) follows the MCP auth spec's scope-selection priority:
1. the `scope` in the server's 401 `WWW-Authenticate` challenge — a new anonymous probe of the MCP endpoint, which also yields the `resource_metadata` URL (tried first for the PRM);
2. else the protected-resource metadata's `scopes_supported`;
3. plus `offline_access` when the AS lists it, so a refresh token is issued;
4. the AS's own list only as a last resort, when the server advertises nothing.

**Existing connections heal on their own.** Discovered scopes now live under `discovered_scopes` and are re-derived on every sign-in (the live challenge is authoritative per spec). A row registered before this split carries its discovery result under `scopes`; it is migrated on the next sign-in and the registered client is reused — no re-creation, no re-registration. `scopes` is reserved for an admin override (`get_oauth_params`: override → discovered).

**Verify runs discovery for DCR** (`_verify_dcr_discovery`): returns `detected {…scopes, scopes_source…}` and states what a sign-in will request, or fails with the reason (undiscoverable metadata / no `registration_endpoint`). Previously Verify passed on reachability alone and the failure only surfaced at Sign in.

**Form** (`MCPConnectionForm.vue`): DCR now has a scopes override field, with `Detected: mcp:read offline_access · from the server's sign-in challenge` under it after Verify (and from `credentials_meta` in edit mode). i18n keys in all ten locales.

## Verification

- **Unit** — `test_mcp_dcr.py` rewritten around an Auth0-shaped fixture (AS advertises OIDC scopes only; the resource scope lives in the PRM and the 401 challenge): selection priority, `offline_access` union, challenge `resource_metadata` first, root/path fallback, open servers, legacy-row migration, admin override precedence, discovery-failure-keeps-existing, `get_oauth_params` fallback. Five new Verify tests in `test_mcp_presets.py`. 248 tests across the OAuth/MCP suites pass.
- **End to end** — a mock MCP server built on the real `mcp` SDK (FastMCP, bearer-protected, `mcp:read` required, PRM at the origin root only, `scope=` in the challenge) behind an Auth0-shaped AS with open DCR, PKCE-checked `/oauth/token` issuing scope-bound tokens. Driven through the running backend: Verify → authorize → consent → callback → token exchange → connection test → tools/list. Sign-in succeeds, token carries `mcp:read offline_access`, tool discovered, credential stored, `credentials_meta` shows the detected scopes.
- **Negative control** — the same run against the previous code reproduces the failure exactly: false-green Verify, OIDC scopes requested, `401 Unauthorized`, credential rolled back.
- **Heal** — a connection registered by the previous code signs in successfully under this code with zero registrations; the stale value is migrated out of `scopes`.
- **Browser** — the form driven with Playwright: field shown for DCR, placeholder before/after Verify, Detected line + source, green result, override reflected as the effective request.
- Frontend production build clean.

https://claude.ai/code/session_01SVv4M3aP1hVUD8DMsVtm9G

---
_Generated by [Claude Code](https://claude.ai/code/session_01SVv4M3aP1hVUD8DMsVtm9G)_


yochze:
**CI status: `unit-tests` was cancelled by the job's 60-minute timeout, not by a failure — and this is a pre-existing `main` condition, not this PR's.**

What the log shows on this head (`0ea5680`): 5312 tests at `-n 4`, steady progress with zero `F`/`E`, reached **97%** at 20:54:21 and was cancelled at exactly 60:00 (`timeout-minutes: 60`).

On `main`: the last green `unit-tests` run (2026-09-14 08:40, `4f2501d`) already took **57m22s**. Since then 13 test files / ~2.3k lines were added under `tests/unit`, and every `e2e tests` run on `main` since (7 consecutive, from 2026-09-14 11:21 through the current head `2b7b172`) has `unit-tests` cancelled at 60 min for the same reason. On the latest `main` run, `e2e-tests (postgres)` was also cancelled at its 120-min ceiling (the last green postgres run took 101 min). This PR adds ~50 unit tests (<1%), all `httpx.MockTransport`-backed and sub-second.

No fix PR exists yet (the workflow file is unchanged). A re-run would time out identically, so I'm not spending one. Proposed patch, separate from this PR since it's CI policy:

```yaml
# .github/workflows/e2e-tests.yml
unit-tests:
  timeout-minutes: 90        # was 60; the suite passed at 57 min two days ago and has grown since
```

The sturdier fix is to shard `tests/unit` across a 2-way matrix (`pytest --dist loadfile` with a split), which halves wall time instead of chasing the ceiling; same idea for the postgres e2e job.

Green on this head: `integration-data-sources`, `integration-llms`, `playwright-tests`. `e2e-tests (sqlite)` and `(postgres)` are still running.

---

PR #1151, opened 2026-09-16
Group agents by stage and add bulk grants to the roles editor

r0533130357-dev:
## Summary

**Agents list (`KnowledgeExplorer`)**
- Agents are grouped under their lifecycle stage — Production, Training, Development, Disabled — with collapsible pill headers.
- Headers render only when the visible agents span more than one stage; a single-stage org sees the flat list as before.
- Headers carry no status dot, so they can't be confused with an agent row's connection-health dot.

**Roles editor (`RolesManager`)**
- Agent access and connection access are two sections rendered from the same markup, so they look and behave identically.
- Each grant is a compact row with an access-level select plus **Custom**, which opens the raw permission checkboxes (replaces the old per-card "Advanced permissions" link):
  - Agents: Can query (`[]`) / Can manage (`manage`)
  - Connections: Can create agents (`create_data_sources`) / Can manage agents (`manage_data_sources`) / Full control (`manage_connection` + `manage_data_sources`)
  - A saved grant that matches no tier renders as Custom.
- **Inline add panel** (no screen change): search, stage filter for agents, select all shown, one access level (including Custom) for every selected resource. Resources already in the role are marked and can't be re-added.
- Agent rows are grouped by stage.
- Nothing is saved until the role's Save; the saved grants are the same per-resource rows as before — no backend or permission-model change. An agent that changes stage keeps its grant.
- Model access is unchanged (it still saves immediately).

**Bug fix:** the editor loaded agents without `show_all`, so private agents the admin isn't a member of were missing from the picker and existing grants on them showed as raw UUIDs. It now requests `show_all=true&include_unconnected=true`. The backend honors `show_all` only for `full_admin_access` / `manage_connections`, so no caller sees more than before.

## Test plan
Verified in a local sandbox (SQLite backend + Nuxt dev, Playwright):
- [x] 6 agents across all stages group correctly; collapse works; "Show all" adds the Disabled group; light and dark mode
- [x] `/data_sources/active` without `show_all` omits a private agent the admin isn't a member of; the editor now lists it by name
- [x] Bulk-added 3 Development agents with Manage + 1 Production agent with Query → saved grants `["manage"]` ×3 and `[]`
- [x] Added an agent with Custom (`manage_instructions`, `manage_evals`) → saved exactly those; reopens as Custom
- [x] Bulk-added 5 connections with "Can manage agents" → saved `["manage_data_sources"]` each; existing grants untouched
- [x] No page errors


r0533130357-dev:
Superseded by #1153, which contains this work (reworked after review) plus agent catalogs, and is based on #1152.

PR #1152, opened 2026-09-17
Pre-release 0.0.563 — umbrella for #1079, #1139, #1147, #1148, #1150

yochze:
Integration branch bundling r0533130357-dev's open PRs on top of `main`, with the small fixes needed to make them work together.

## Bundled PRs (merged as-is)

| PR | What |
|---|---|
| #1002 | Docker: auto-generate and persist `BOW_ENCRYPTION_KEY` (env var → keyfile → generate, before the workers fork) |
| #1079 | Artifact version management: parent identity table + per-artifact version chains (migration `artver01`) |
| #1139 | Browse and preview file connections in the agent Files tab |
| #1147 | RTL: directional-icon auto-flip matches the emitted `i-heroicons:*` class |
| #1148 | Keep the agent Files tree and the Uploaded panel in sync |
| #1150 | Indexing liveness: heartbeat, stale-run reaping, stage timeouts (migration `idxactivity01`) |
| #1155 | Scoped instruction search (`GET /api/knowledge/search?data_source_id=`) + clear button on "Search everything" — frontend reworked, see below |
| #1156 | Monitoring "All time" means the org's whole history (Explore, Cost, Diagnosis); Diagnosis gets an "All time" preset with monthly buckets |
| #1157 | Security: `@nuxtjs/mdc` 0.17 → 0.22 (Snyk high SNYK-JS-NUXTJSMDC-19883915); `package.json` + `yarn.lock` only |

Every one of those branches has been re-merged and reports "Already up to date", so nothing on them is outstanding. `main` is merged in as well, so #1149 (MCP DCR scopes) is included via its merge.

Also on the branch: 408f58f "Rename custom queries to custom tables and add Power BI (DAX) custom tables" (merged in from a sibling session; no migration).

Deliberately left out: #1080 (superseded by the Diagnosis explorer that landed on main), #1081 (scheduled task templates), #1151 (agent stage grouping), #1153 (agent catalogs).

## Integration commits on top

- **Port the browser verification policy to the split artifact tables.** The artifact verification policy on main queried the `Artifact` model by the id the API hands out. Under #1079 that id is a version id and `Artifact` is the parent table, so every browser tool was denied (8 failures in `test_artifact_browser_policy.py`). The policy now queries `ArtifactVersion`. #1079 will hit the same break the moment it is rebased.
- **Re-chain `idxactivity01` onto `artver01`.** #1150's migration was written against `codeperm01`, the same parent as `artver01`; merged together alembic sees two heads and every test errors. One-line change so the chain stays linear.
- **Instructions open in the right pane (rework of #1155's UI).** #1155 put a search row inside each agent's Instructions group in the tree, which swapped the folder tree for a flat hit list in the narrow sidebar. Instead, clicking an agent's Instructions row now opens an Instructions panel in the right pane, like Tables / Tools / Files: the agent's full list, a search box wired to #1155's scoped endpoint (matches bodies, never global or other agents' rows), a clear button, Esc to clear, count, Add. Rows are the tree's `InstrLeaf` plus a one-line body snippet so a hit shows why it matched; the tree's filter popover applies to the panel too; each agent keeps its own query, re-run on reopen. The agent overview's "N instructions" count opens the panel (mobile as well). Backend, tests and locale strings from #1155 are unchanged.
- **Reconcile #1002 with the ephemeral-key guard main has added since.** Two conflicts, both in encryption-key resolution. `BowConfig.validate_encryption_key` takes #1002's order (explicit YAML literal → `BOW_ENCRYPTION_KEY` env var → generate), which is what the surrounding comment already promised, and keeps main's `mark_encryption_key_ephemeral()` on the generate path so payload encryption stays off for a key only this process can read. `config.resolve_env_vars` takes #1002's side: the loader no longer mints a key of its own for an unresolved `${BOW_ENCRYPTION_KEY}`, it returns `None` and the validator above handles env-var-then-generate, still flagging ephemeral.
- **Two CI-found robustness fixes.** `test_field_predicates` (Diagnosis explorer) pinned its reference time at import while the query is compiled against the request clock, so it failed whenever it ran late in the two-hour postgres leg; its fixture row now sits 12h inside the window. The network share client's listing failed the whole call when a file vanished between the walk and the stat (a sibling worker's temp file under `/tmp`); it now skips such entries.
- **CI: `unit-tests` timeout raised from 60 to 120 minutes.** The suite takes about an hour on a hosted runner, so the job was cancelled before reporting on every PR.
- Version bump to 0.0.563 with changelog.

## Verification

- Backend unit suite: 5239 passed. Remaining failures are environmental (proxy CA bundle for the SharePoint on-prem tests, missing `libodbc` for Sybase) and pass with those cleared.
- E2E on SQLite: artifact browser policy 10/10, RBAC artifact suites 47/47, file browser, artifact, doc, console, scheduled prompt suites green; indexing liveness 7/7; migration chain replay and `artver01` backfill round-trip green; `test_knowledge_search_scope.py` + `test_instruction_light_view.py` 9/9; after #1156, `test_console_metrics.py` + `test_diagnosis_explorer.py` 34/34 (includes its two new all-time tests and our timing fix).
- After #1002: `test_bow_config_encryption_key.py` + the three encrypted-payload suites 51/51. All three resolution paths probed directly — an explicit YAML literal wins over the env var, the env var wins for an omitted / empty / null / `${…}` key, and with nothing set a valid per-process Fernet key is generated, warned about, and flagged ephemeral (so payload encryption stays off). The real `start.sh` block was exercised in a temp volume: first boot generates and persists 0600 with no trailing newline, a restart reloads the identical key, an explicit env key wins and leaves the keyfile untouched, an unwritable keyfile falls back to a temporary key with the loud warning, an empty keyfile is treated as missing and regenerated, and a malformed explicit key exits 1. The sandbox backend then rebooted on the merged code with its pinned key: healthy, no ephemeral warning, and the stored Anthropic key and data-source credentials still decrypt.
- Frontend: `yarn install --frozen-lockfile` on #1157's lockfile resolves `@nuxtjs/mdc` 0.22.2 / shiki 4.4.3; production build passes on the final head.
- Live sandbox with Claude 4.5 Haiku, driven through Playwright: dashboard create → edit (v2) → separate dashboard (own artifact, v1) → doc (v1) → rebuild from scratch (v3 of the same artifact) → "Use this version" (v4), per-mode hide rule, chain-wide rename; real alembic downgrade → legacy flat rows → upgrade → downgrade → upgrade round trip with legacy grouping, legacy revert and thumbnail-row revert in the UI; file browser previews for CSV, XLSX (sheets, row cap), PDF, DOCX/PPTX via LibreOffice, images, animated images frozen, oversize refused, scope/escape 403s with audit rows; tree↔panel upload/delete sync; Hebrew RTL icon flips; MCP DCR against a mock Auth0-shaped IdP (scopes from the 401 challenge, PKCE, tools discovered); orphaned indexing run reaped and reindex allowed; stuck-run warning shown after 5 minutes without activity.
- Instructions pane, same sandbox: list of 7 (published, draft, archived rows styled as in the tree) → "revenue" returns the 2 rows whose bodies mention it (one has no "revenue" in its title) → nonsense query shows "No results." → ✕ and Esc restore the list → clicking a hit opens the instruction and closes the pane → reopening remembers the query → tree filter "Inactive" narrows the pane to the draft row → Finance agent's search returns only its own hit (not Chinook's, not the global one) → overview "7 instructions" opens the pane → mobile: full-width pane, no keyboard auto-focus, Back returns to the tree → Hebrew: mirrored layout with translated placeholder, clear label and empty state.
- Monitoring, same sandbox, after #1156: Diagnosis `?range=all` shows the "All time" preset, "6 runs match of 6 in All time", and its request carries no start/end; Explore's "All Time" period sends only an end date and shows the org's totals. After #1157: a report's chat messages still render through `<MDC>` (lists, bold, paragraphs) with no page or console errors.

https://claude.ai/code/session_01MkwkiyPoyMYKXb1z1rART4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -221,7 +221,7 @@ RUN mkdir -p /app/certs && \
 # These paths match volume mounts in docker-compose.yaml; they must exist with
 # app-user ownership so Docker seeds named volumes with writable perms on first run.
 RUN mkdir -p /app/backend/uploads/files /app/backend/uploads/branding \
-             /app/backend/branding_uploads /app/backend/logs && \
+             /app/backend/branding_uploads /app/backend/logs /app/backend/data && \
     chown -R app:app /app
 
 WORKDIR /app
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -12,7 +12,9 @@
 #   POSTGRES_DB=bagofwords               # Database name
 #   POSTGRES_PORT=5432                   # PostgreSQL port (optional)
 #   APP_PORT=3000                        # Application port (optional)
-#   BOW_ENCRYPTION_KEY=                  # Fernet encryption key (44 chars, ends with =)
+#   BOW_ENCRYPTION_KEY=                  # Optional: Fernet key (44 chars, ends with =).
+#                                        # If unset, a key is auto-generated on first
+#                                        # start and persisted in the app_data_dev volume.
 #   BOW_LICENSE_KEY=                     # Enterprise license key (optional)
 #
 # Generate encryption key:
@@ -65,6 +67,8 @@ services:
       - uploads_data_dev:/app/backend/uploads
       - branding_data_dev:/app/backend/branding_uploads
       - logs_data_dev:/app/backend/logs
+      # Persists the auto-generated encryption key across restarts
+      - app_data_dev:/app/backend/data
     depends_on:
       postgres:
         condition: service_healthy
@@ -85,6 +89,8 @@ volumes:
   branding_data_dev:
     driver: local
   logs_data_dev:
+    driver: local
+  app_data_dev:
     driver: local
 
 networks:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,7 +12,9 @@
 #   POSTGRES_USER=bow                     # PostgreSQL username
 #   POSTGRES_PASSWORD=your_secure_pw      # PostgreSQL password (CHANGE THIS!)
 #   POSTGRES_DB=bagofwords                # Database name
-#   BOW_ENCRYPTION_KEY=                   # Fernet encryption key (44 chars, ends with =)
+#   BOW_ENCRYPTION_KEY=                   # Optional: Fernet key (44 chars, ends with =).
+#                                         # If unset, a key is auto-generated on first
+#                                         # start and persisted in the app_data volume.
 #   BOW_LICENSE_KEY=                      # Enterprise license key (optional)
 #
 # Generate encryption key:
@@ -68,6 +70,8 @@ services:
       - uploads_data:/app/backend/uploads
       - branding_data:/app/backend/branding_uploads
       - logs_data:/app/backend/logs
+      # Persists the auto-generated encryption key across restarts
+      - app_data:/app/backend/data
     depends_on:
       postgres:
         condition: service_healthy
@@ -107,6 +111,8 @@ volumes:
   branding_data:
     driver: local
   logs_data:
+    driver: local
+  app_data:
     driver: local
   caddy_data:
     driver: local
--- a/start.sh
+++ b/start.sh
@@ -3,12 +3,30 @@
 # Set environment variables
 export ENVIRONMENT=production
 
-# Generate BOW_ENCRYPTION_KEY if not provided (must happen BEFORE workers fork)
+# Resolve BOW_ENCRYPTION_KEY (must happen BEFORE workers fork so every worker
+# shares the same key). Resolution order:
+#   1. BOW_ENCRYPTION_KEY env var (explicit key management: k8s Secrets, vaults)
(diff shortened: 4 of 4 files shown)


yochze:
CI status note on `unit-tests`: the job shows **cancelled** on every push to this PR, and on every other recent PR in the repo (#1079, #1139, #1147, #1148, #1150, #1151). It is the workflow's own `timeout-minutes: 60` on the `unit-tests` job in `.github/workflows/e2e-tests.yml`; the suite takes ~58 minutes locally at `-n 4` and the hosted runner is slower, so it never reaches the end. Not caused by this branch and not fixable inside it without changing CI config, so I'm leaving the workflow alone here.

For the record, the full unit suite was run locally on this branch: 5239 passed; the remaining failures are environmental (proxy CA bundle for the SharePoint on-prem tests, missing `libodbc` for Sybase) and pass with those cleared. The one real e2e failure seen on this PR (`test_field_predicates` on the postgres leg) was a timing hazard in the test itself and is fixed in 8a4d7f7.

---

PR #1158, opened 2026-09-19
fix(llm): price Claude Fable 5.1 / Mythos 5.1 cache reads at 0.025x, not the family 0.1x

yochze:
Claude Fable 5.1 and Claude Mythos 5.1 bill a cache hit at **0.025x** base input — $0.25/MTok against their $10 base — where the rest of the Claude family pays 0.1x. `app/ai/llm/pricing.py` sets one rate for the whole family, so every cached read on those two is costed at **4x** its real price.

**This PR has changed shape since it was opened — read this section before the rest.** It began as "add Claude Fable 5.1 to the catalog". That catalog row is **already on `main`** (merged as `2111c93`, released in 0.0.564), so it is no longer part of this diff. Separately, #1159 (`16e642d`) rewrote cached-token pricing, moving the arithmetic out of `LLMUsageRecorderService._calc_input_cost` into the new family-and-TTL-keyed `pricing.py`. This branch took that refactor wholesale and deleted its own earlier version of the fix. What remains here is **only the per-model cache-read rate**, re-landed where the family rates now live.

The bug survived #1159's refactor — and got wider. `pricing.py` hardcodes `ANTHROPIC: CacheRates(read=0.10, …)`, and because that module deliberately prices on model family rather than provider account, the 4x overcharge now applies on every route it covers (first-party, Vertex, Bedrock, Azure), not just the first-party one.

## Why a customer cares

An admin who moves their organization onto Claude Fable 5.1 reads their spend off the Cost console. Cache reads dominate a long agent run — one report conversation replays a cached system prompt, schema context and transcript on every turn — and #1159 just made caching hold *better* across a run, which raises the cached share and so the size of this error. On the most expensive model we serve, the number they budget against is materially and silently high. Nothing in the UI would catch it: the figure is plausible, internally consistent, and wrong in the safe-looking direction. Finance plans against it; an admin comparing Fable 5.1 against Opus 5 makes the wrong call on a real cost difference.

## Product impact — what the admin sees

A report conversation on Fable 5.1 with 2M cached-read tokens and 100k output tokens:

| Line item | Billed by Anthropic | Reported before | Reported after |
|---|---|---|---|
| 2M cache reads | $0.50 | $2.00 🔴 | $0.50 🟢 |
| 100k output | $5.00 | $5.00 | $5.00 |
| **Conversation total** | **$5.50** | **$7.00 🔴 (+27%)** | **$5.50 🟢** |

The skew grows with the cache ratio: a cache-heavy run approaches the full 4x overstatement on the cached portion.

## Technical mechanism

```mermaid
flowchart TD
    A["rates_for(provider_type, model_id)"] --> B["family = resolve_family(...)"]
    B --> C["BEFORE: return _RATES[family]<br/>ANTHROPIC.read = 0.10 for everyone"]
    C --> D["claude-fable-5-1 @ $10 base<br/>1M cache reads → $1.00 🔴<br/>true price $0.25"]
    B --> E{"AFTER: family == ANTHROPIC<br/>and id matches an override?"}
    E -->|"fable-5-1 / mythos-5-1"| F["replace(rates, read=0.025)<br/>→ $0.25 🟢"]
    E -->|no match| G["family rate unchanged:<br/>Fable 5, Opus 5, Sonnet 5, Haiku 🟢"]
    E -->|"OpenAI family"| H["never overridden —<br/>read means 'cached price'<br/>behind a rebate there 🟢"]
    style C fill:#ffd7d5
    style D fill:#ffd7d5
    style F fill:#d4f8d4
    style G fill:#d4f8d4
    style H fill:#d4f8d4
```

Two deliberate constraints on the override:

- **Version-qualified tags, never the family.** `claude-fable-5` and `claude-mythos-5` genuinely are 0.1x, so a bare `fable-5` / `mythos` match — the substring style used in `anthropic_client.py` — would sweep them in and *under*-report their cost by 4x, the same bug reversed. An id matching nothing keeps its family rate, which is the safe direction for a model nobody has priced yet.
- **Anthropic family only.** `read` is an additive cost for the Anthropic family but the *cached price* behind a rebate for the OpenAI family, where the math is `cost -= reads × per_token × (1.0 − read)`. A 0.025 leaking across would quietly compute a 97.5% refund. There is a test for exactly this.

Cache **writes** are untouched: Anthropic publishes the same 1.25x / 2x for these models, which is what the family already carries.

## Reviewer decision box

| | |
|---|---|
| **Risk** | Low. One exception list and one `replace()` in `rates_for`; no call sites changed. |
| **Blast radius** | The cached-read term of every LLM usage record. Numerically **unchanged for every model except the two named ids** — which no organization can be running today, since `claude-fable-5-1` only reached the catalog in 0.0.564 and `claude-mythos-5-1` is not in the catalog at all (it is reachable only as a custom model). No schema change, no migration, no API shape change. |
| **Behavior change (intended)** | ✅ Fable 5.1 / Mythos 5.1 cache reads cost 0.025x input instead of 0.1x. Historical `llm_usage_record` rows are not rewritten. |
| **Verified** ✅ | 343 LLM/pricing/usage unit tests and 39 e2e tests green on the merged tree. #1159's own 30 cache-pricing tests still pass unchanged — the override does not disturb the family rates they assert. New tests proven to fail (5 cases) against the pre-fix `rates_for`. Rates re-checked against Anthropic's live pricing page today. |
| **NOT verified** ⚠️ | No live API round-trip against `claude-fable-5-1` (no Anthropic key in this environment), so `tests/integrations/llm_clients.py` and the "Test connection" UI check are **unrun**. No browser/UI check. Full `tests/unit` (~60 min) not run — only the LLM/pricing/usage selection. The 0.025x rate is taken from Anthropic's first-party pricing page and applied on every route, following `pricing.py`'s own premise that Claude prices identically everywhere; **partner-operated Bedrock and Vertex publish their own pricing and were not separately checked.** |
| **Where to look first** | `backend/app/ai/llm/pricing.py` — `_ANTHROPIC_READ_RATE_OVERRIDES` and the family gate in `rates_for`. Confirm `fable-5-1` cannot match `claude-fable-5`, and that a future Fable version won't silently inherit the reduced rate. |

## Evidence

| Check | Result | Attributable to this PR |
|---|---|---|
| `tests/unit -k "llm or vertex or pricing or cache or usage or cost or image_generation"` (merged tree) | **343 passed**, 0 failed | — |
| LLM catalog + usage e2e (4 files, merged tree) | **39 passed**, 0 failed | — |
| `tests/unit/test_cached_token_pricing.py` (#1159's file, now carrying these cases) | **44 passed** | 14 new |
| Same file vs. pre-fix `rates_for` | **5 failed**, 39 passed | proves the new tests can fail (tests/AGENTS.md rule 6) |
| `ruff check` on changed files | 16 `UP045`, of which **14 pre-exist on `main`** | the 2 new ones are `Optional[...]` annotations matching the idiom `pricing.py` already uses throughout; converting them alone would leave the module inconsistent |

No pre-existing test failures were encountered, so there is no attribution appendix.

## Verification against official sources

[Pricing](https://platform.claude.com/docs/en/about-claude/pricing), model-pricing table and its footnote:

> Cache hits and refreshes on Claude Fable 5.1 and Claude Mythos 5.1 are priced at 0.025x the base input price. All other models use the standard 0.1x multiplier.

The same table lists both at $10/MTok base with 5m writes at $12.50 and 1h writes at $20 — i.e. the family's 1.25x / 2x, which is why only `read` is overridden. Claude Fable 5 and Claude Mythos 5 are listed at $1/MTok cache hits, confirming they stay at 0.1x.

## Context map

**Key files**
- `backend/app/ai/llm/pricing.py` — `_ANTHROPIC_READ_RATE_OVERRIDES`, `_anthropic_read_rate_override`, and the family-gated override in `rates_for`.
- `backend/tests/unit/test_cached_token_pricing.py` — the per-model-rate section appended to #1159's file.

**Convention this PR establishes:** a model whose cache rate departs from its family's is added to the exception list in `pricing.py`, next to the family rates so the two cannot drift, matched on a version-qualified tag and defaulting to the family rate when nothing matches. Not a new branch at the call site, and not a second constant elsewhere.

**Commits by role**
- `3b3a4d7` — the original catalog row; **already on `main` via `2111c93`**, listed here only because it is still in this branch's history.
- `eae1f9e` — the first version of this fix, against the pre-#1159 recorder. Superseded.
- `b0c916f` — merge of `main` + the fix re-landed in `pricing.py`, tests folded into #1159's file, VERSION → 0.0.565.

**Follow-ups deliberately left out of scope**
- **Claude Sonnet 5 is mispriced in the catalog.** `LLM_MODEL_DETAILS` carries $3/$15 per MTok, but Anthropic's pricing page says the $2/$10 introductory rate *became* the standard price and that "the previously scheduled increase to $3/$15 per million input/output tokens on September 1, 2026 will not occur." Sonnet 5 is the **default** Anthropic model, so every organization on the default is seeing its spend over-reported by 50% — a larger customer impact than the bug this PR fixes. One-line data change, but it moves a visible number for every existing organization, so it wants its own PR and release note rather than riding along here.
- The `add-llm-provider-or-model` skill's verification snippet asserts `len(ids) == len(set(ids))` over `LLM_MODEL_DETAILS`. That assert **already fails on `main`**: `claude-sonnet-5` and `gemini-3.1-pro-preview` are each listed twice on purpose, once first-party and once for `vertex`. The snippet predates the Vertex entries and should key on `(model_id, provider_type)`.
- Fable 5.1 is not in the `vertex` catalog subset. Google Cloud serves `claude-fable-5-1`, but that subset is deliberately minimal, so it belongs in its own change.


yochze:
### `unit-tests` is red on this head, but the failure is not this PR's

**Failing:** [`unit-tests`](https://github.com/bagofwords1/bagofwords/actions/runs/35461853368/job/105947016534) — `1 failed, 5412 passed, 20 skipped`.

```
FAILED tests/unit/test_sandbox_feedback_loop.py::TestSandboxRulesInPrompts::test_create_data_prompt_lists_forbidden_builtins
AssertionError: create_data prompt missing sandbox rules for:
['__import__', 'breakpoint', 'bytearray', 'compile', 'delattr', 'eval', 'exec',
 'exit', 'getattr', 'globals', 'hasattr', 'input', 'locals', 'memoryview',
 'open', 'quit', 'setattr', 'vars']
```

**Why it isn't this PR's.** This branch changes `app/ai/llm/pricing.py`, its test, `CHANGELOG.md` and `VERSION` — nothing on the codegen path. The failure reproduces identically on pristine `origin/main` (`eb6c739`), checked out in a clean worktree with this branch nowhere in it:

```
$ git worktree add /tmp/mainwt origin/main && cd /tmp/mainwt/backend
$ pytest "tests/unit/test_sandbox_feedback_loop.py::TestSandboxRulesInPrompts"
1 failed, 1 passed in 6.50s   # same assertion, same 18 builtins
```

It reproduces in ~7 seconds and is fully deterministic, so I confirmed it by running it on the base branch rather than by spending another hour-long CI re-run; that is the stronger evidence for the same question. `unit-tests` is also `continue-on-error: true` in `e2e-tests.yml`, so it is not gating this PR either way.

### Root cause — a stale test, **not** a dropped sandbox rule

Worth stating plainly, because the assertion text reads alarmingly: **the sandbox rules are still sent to the model in full.** Nothing is missing from the prompt.

#1159 (`c727020`) split `Coder.generate_code` into a run-invariant system half and a per-call user half so the system prefix can carry a cache breakpoint (`coder.py:782-798`, sent as `system=system_text` at `coder.py:971-972`). `_sandbox_rules_section()` moved into that system half. The test's stub only ever records the user half:

```python
class _PromptCapturingLLM:
    async def inference_stream_v2(self, messages, **kwargs):
        self.prompts.append(messages[0].content)   # user half only; `system` dropped
```

Instrumenting the real call on `main` confirms where the rules went:

```
system half present: True | system chars: 19835
missing from USER half only : 18
missing from SYSTEM half    : 0
missing from BOTH COMBINED  : []
```

`test_inspection_prompt_lists_forbidden_builtins` still passes because `generate_inspection_code` was not split — it keeps `_sandbox_rules_section()` inline in the single user message (`coder.py:1079`). That asymmetry is exactly the tell.

### Proposed patch (not applied here — it would widen a pricing PR)

The test asserts prompt *layout* when the invariant it means to protect is "the model is told the sandbox rules." Capturing both halves restores that and makes it indifferent to future prompt splits:

```diff
 class _PromptCapturingLLM:
     """Stands in for Coder.llm — records the rendered prompt."""

     def __init__(self):
         self.prompts = []
+        self.systems = []

     async def inference_stream_v2(self, messages, **kwargs):
         self.prompts.append(messages[0].content)
+        self.systems.append(kwargs.get("system") or "")
         yield TextDeltaEvent(text=CLEAN_CODE)
+
+    def rendered(self, i: int = 0) -> str:
+        """Everything the model actually sees for call i, both halves.
+
+        generate_code splits its prompt into a cacheable system half and a
+        user half; generate_inspection_code does not. Which half carries a
+        rule is a caching decision, so assertions about what the model was
+        told read both.
+        """
+        return f"{self.prompts[i]}\n{self.systems[i]}"
```

then in both `TestSandboxRulesInPrompts` tests:

```diff
-        prompt = coder.llm.prompts[0]
+        prompt = coder.llm.rendered()
```

Applying it to both sandbox tests, not just the failing one, keeps them symmetric and stops the inspection twin breaking the day that path gets the same caching treatment.

I have not pushed this: it is unrelated to this PR's diff, and no fix for it exists yet that I could port. Happy to send it as its own one-file PR against `main` — it affects every PR's `unit-tests` run, not just this one.

---

PR #1181, opened 2026-09-23
Fix transcript rendering mutations, artifact caching, and default reasoning effort

yochze:
## Summary
This PR fixes critical issues where rendering prompts was mutating the live transcript and where artifact edits were invalidating the entire cached prefix. It also improves token budgeting to account for tool call arguments, and stops reasoning "off" from running Sonnet 5 / Opus 4.7+ / Fable 5 at the provider's default (high) effort.

## Key Changes

- **Prevent transcript mutations during rendering**: Added `_detached()` function to create deep copies of transcript turns before rendering. This prevents the volatile head (conversation history, current artifact) from accumulating in the live transcript across iterations.

- **Move artifact rendering to volatile head**: Moved `_render_current_artifact()` from the static context block to `_build_turn_head()`. This ensures artifact edits (which change id, version, and code) no longer invalidate the cached prefix of the entire transcript, reducing token re-computation from hundreds of thousands to just the artifact's size.

- **Fix token estimation for prompts**: Updated `build_prompt()` to include all messages, not just the first one. The volatile head now rides on the last turn, so reading only `messages[0]` was under-counting tokens.

- **Account for tool call arguments in budget**: Modified `Turn.tokens()` to include the serialized arguments of `ToolCallPart` objects. Tool call arguments (especially generated code in `create_data`/`edit_artifact`) are replayed every iteration and can be tens of thousands of characters, so they must be visible to the budget calculation.

- **Reasoning "off" → low effort on always-thinking models** (commit 564a311): these models cannot turn thinking off, and a request with no effort ran at the provider default (high). The Anthropic client now sends `output_config.effort = "low"` when the caller asked for none. Trigger phrases ("think hard", …) still resolve to high; per-completion and model-default settings pass through unchanged; Haiku 4.5 / Sonnet 4.6 and older still send no thinking.

## Implementation Details

- The `_detached()` function performs shallow copies of turns and their parts, preserving any decay applied by the loop while preventing rendering mutations from affecting the live transcript.
- Artifact rendering now happens per-turn in the volatile head rather than once in the static context, allowing the cache prefix to remain stable across artifact edits.
- Token counting for `ToolCallPart` arguments uses `estimate_tokens()` on the JSON-serialized args, consistent with how other content is counted.
- The low-effort default lives in the Anthropic client (`OFF_EFFORT_FOR_ALWAYS_THINKING` in `reasoning.py`), not in `_effort_to_thinking_config`: the Bedrock client converts any thinking config to `budget_tokens`, which Sonnet 5 rejects, so "off" must keep resolving to no config there.
- `test_summary_requested_without_effort_override[anthropic]` encoded the old "never set an effort" rule; renamed to `test_summary_requested_when_no_effort_given` and now asserts `low`.
- Comprehensive test coverage added to verify: history doesn't accumulate, render-time decay doesn't stick, loop-applied decay is preserved, artifact edits don't break caching, token estimates include volatile content, and the effort sent per model / trigger / explicit setting.

## Verification (live sandbox, real Anthropic API)

| | Before | After |
|---|---|---|
| History copies in the last request of a 9-step turn (Haiku 4.5) | 8 | 1 |
| Cache re-writes in that turn | 81k tokens | 17k tokens |
| Largest single cache re-write | 35.8k | 4.6k |
| Sonnet 5, 3-turn scenario: reasoning time | 33.6s | 14.4s |
| Sonnet 5: output tokens | 8.1k | 6.1k |
| Sonnet 5: run time | 119s | 102s |

Same completed dashboard in both Sonnet 5 runs (3 artifact versions, all KPIs present); a "think hard" turn still sent effort `high`. One run per side, so treat the timings as indicative. Trade-off from the artifact move: uncached input per step rises by roughly the artifact block's size.

https://claude.ai/code/session_01HkYGC388Rt52CSeDyCVYKR


yochze:
`unit-tests` is red on 8226ad3 with 7 failures. All 7 also fail on `main`: same tests, same assertions, in [main's run on 8683167](https://github.com/bagofwords1/bagofwords/actions/runs/35880699678/job/107248081729) (`7 failed, 5538 passed`). None of them tests code this PR changes:

- `test_opus_5_5_support.py::test_opus_5_family_never_gets_budget_tokens` (6 cases): the test expects `{"type": "adaptive"}`, but `_effort_to_thinking_config` on main returns `{"type": "adaptive", "effort": <level>}`. This PR doesn't touch that function.
- `test_sandbox_feedback_loop.py::test_create_data_prompt_lists_forbidden_builtins`: the create_data prompt no longer lists the forbidden builtins.

The 5,547 other unit tests pass, including every test this PR adds. I found no open PR fixing these, so there's nothing to port yet. I'm not re-running: the failures reproduce identically on main and locally, and CI is already running on the new head (564a311). Both need a fix on main (either update the tests or restore the prompt text), which is outside this PR's scope.

---

yochze:
`e2e-tests (postgres)` on 564a311 was **cancelled by the 120-minute job timeout**, not failed. The log ends mid-run (`The operation was canceled` during a test's setup), and no test failure appears before the cutoff.

It isn't this PR's change: the suite runs at the edge of the limit everywhere.

| Run | "Run E2E tests" step |
|---|---|
| `main` @ 8683167 | 113.8 min (passed) |
| this PR @ 8226ad3 | 112.6 min (passed) |
| this PR @ 564a311 | >118 min, killed at 120 |

564a311 only changes the thinking/effort fields the Anthropic client puts on its request, which e2e never waits on. The two runs of this PR also overlapped for about 1.5h.

I tried the one allowed re-run (`rerun-failed-jobs`), but it was refused (403: this integration can't re-run workflows). **Could someone with Actions access re-run the failed jobs?**

The durable fix is outside this PR. Raise `timeout-minutes` for the `e2e-tests` job in `.github/workflows/e2e-tests.yml` (currently 120, line 17), or split the Postgres suite, since main already uses ~95% of the budget.

The `unit-tests` failures on this head are the same 7 already red on `main` (see my previous comment); there are no new failures.

---

PR #1182, opened 2026-09-23
fix(agent): stop "greenlet_spawn" failures in parallel create_data and unblock parallel tool persistence

yochze:
**What:** coder-reasoning snapshots no longer write through the agent's shared DB session. On Postgres, parallel tools no longer queue on a lock their writers don't need.
**Why:** a snapshot commit that collided with a sibling tool on Postgres rolled back the shared session. That expired every loaded object, and the next read raised `MissingGreenlet`, failing the whole run. Separately, `_tool_db_lock` serialized each tool's step/visualization persistence even though those writers use their own sessions.

## Executive summary
A customer on Postgres asked for several tables at once and got **"Agent failed: greenlet_spawn has not been called"**. The planner had run three `create_data` calls in parallel, and the new reasoning stream (from `a7a8ca7`) crashed the run. After this PR:
- that failure path is gone;
- multi-table requests on Postgres also finish sooner, because tools no longer wait for each other to save results (parallel batch 78s → 43s average in our sandbox).

SQLite deployments behave as before.

## Visuals
**4a. Product impact**
```mermaid
flowchart LR
  U[User asks for 3 tables] --> P[Planner runs 3 create_data in parallel]
  P -->|before| F["Run dies: Agent failed: greenlet_spawn..."]:::bad
  P -->|before| S[Tables finish ~13s apart, one after another]:::bad
  P -->|after| OK[All 3 tables; batch ~45% faster on Postgres]:::good
  classDef bad fill:#fdd,stroke:#c00
  classDef good fill:#dfd,stroke:#080
```
**4b. Mechanism**
```mermaid
flowchart TB
  subgraph Before
    R1[reasoning snapshot] -->|commit on shared self.db| C1{sibling tool mid-query?}:::bad
    C1 -->|asyncpg: another operation in progress| RB[self.db.rollback expires ALL objects]:::bad
    RB --> MG[next_seq reads expired latest_seq -> MissingGreenlet, every event and retry]:::bad
  end
  subgraph After
    R2[reasoning snapshot] -->|Postgres: own short-lived session| OK2[shared session untouched]:::good
    R2 -->|SQLite single-writer: self.db under _tool_db_lock| OK3[single writer kept]:::good
    SEQ[next_seq: unmapped counter] --> OK4[survives rollbacks, never lazy-loads]:::good
  end
  classDef bad fill:#fdd,stroke:#c00
  classDef good fill:#dfd,stroke:#080
```

## Reviewer decision box
| | |
|---|---|
| **Risk** | Medium. It touches shared-session concurrency in `agent_v2.py`. Each part is verified on both backends in real UI runs. |
| **Blast radius** | Agent runs that stream coder reasoning (`create_data` / `inspect_data` / `write_csv` with thinking), and parallel tool batches on Postgres. |
| **Intended behaviour changes** | Postgres only: `_handle_tool_output` / `_handle_streaming_event` run without `_tool_db_lock`. Table-usage analytics are written in the background, drained before `completion.finished`. SQLite is unchanged. |
| **Verified ✅** | Deterministic Postgres repro (the customer's exact error); real UI runs before/after on Postgres (3 scenarios, 12 runs) and SQLite; event→fetch consistency; unit tests. |
| **NOT verified ⚠️** | The customer's environment. The speed-up on a same-host Postgres: the timings used a 15ms-latency proxy, so the gain will be smaller there. `OPENAI_API_KEY_TEST`-gated tests (key not available here). |
| **Look first** | `_coder_reasoning_callback` (single-writer vs own session), `_single_writer_guard` and its two call sites, `_record_step_table_usage_bg`, `ProjectManager.next_seq`. |

## Evidence
| Check | Result |
|---|---|
| Repro script, real Postgres/asyncpg (`tools/agent/repro_reasoning_greenlet.py`) | before: 58× `MissingGreenlet: greenlet_spawn has not been called`; after: 0 errors in every mode |
| Unit tests (`test_stream_session_isolation.py`) | 8 new cases, all pass; they fail on the unfixed code. Related suites: **111 passed** |
| Postgres, 3 tables in parallel, UI-driven, Haiku, 3 runs each | batch 82/76/76s → 44/47/37s; whole run 117/108/111s → 77/81/72s |
| Postgres, same with data source attached, 2 runs each | batch 94/94s → 48/41s; table usage recorded 7/7 and 8/8 → 7/7 and 7/7 |
| `_tool_db_lock` held / waited per 3-tool run | 61–82s / 72–96s → 6–7s / 7–9s |
| Event→fetch (fetch step/visualization right after each event, as the UI does) | 0 mismatches before (36 checks), 0 after (36 checks) |
| SQLite, same parallel scenario | `40af090` success in 42s; **`2951f80` timed out (self-introduced, see below)**; this PR success in 40s |
| **Failures attributable to this PR** | **None open.** One self-introduced SQLite regression was found and fixed in `da55b20`, with a regression test. |

Present before and after, not caused by this PR: duplicate SSE `seq` values from `_handle_streaming_event`'s fresh-copy `next_seq`, identical in proportion; and the `inspect_data` table-usage `'NoneType'` warning.

## Context map
- **Key files:** `backend/app/ai/agent_v2.py`, `backend/app/project_manager.py` (`next_seq`), `backend/tests/unit/test_stream_session_isolation.py`, `tools/agent/repro_reasoning_greenlet.py`, `tools/agent/pg_latency_proxy.py`.
- **Convention:**
  - Streaming and best-effort writers never commit or roll back `self.db` off single-writer mode; use a short-lived session.
  - In single-writer mode (SQLite), writes stay on `self.db` under `_tool_db_lock`.
  - Take `_tool_db_lock` only around code that actually touches `self.db`; use `_single_writer_guard()` for writers that go through `_writes_session()`.
- **Commits:**
  - `2951f80`: greenlet fix (reasoning on its own session, rollback-proof seq counter, crash-handler reload).
  - `da55b20`: fixes the SQLite regression that `2951f80` introduced.
  - `9799cc7`: conditional lock and background table usage.
- **Reports:** `docs/feedback-loops/coder-reasoning-missing-greenlet.md`, `docs/feedback-loops/parallel-tool-lock-and-table-usage.md`.
- **Follow-ups (not in this PR):**
  - `create_data` / `inspect_data` read the shared session without the lock (loadables, snippets, viz instructions, the `inspect_data` excerpt).
  - `refresh()` after every commit reloads the report graph (~270 statements per `create_data`).
  - `refresh_warm` runs 10–11× per run.
  - Duplicate SSE `seq` values.
  - The `mark_error_on_latest_block` multi-row bug.

## Deep detail and honesty
- **Root cause:** `a7a8ca7` saved coder reasoning through the shared `AsyncSession` while parallel siblings used it. asyncpg rejects concurrent operations on one connection; SQLite's driver queues them, which is why only Postgres failed. The resulting rollback expired all instances. `next_seq` then lazily loaded `latest_seq` → `MissingGreenlet`. The crash handler only reloaded objects when `not db.is_active`, so every retry failed the same way.
- **Self-introduced regression, disclosed:**
  - `2951f80` moved reasoning saves to their own session everywhere. On SQLite the agent's session is the only writer and holds the lock through a tool run, so every snapshot waited out `busy_timeout` inside the coder stream, and parallel `create_data` hit the 300s hard timeout.
  - Caught by the SQLite real-scenario run, confirmed against `40af090`, and fixed in `da55b20`.
  - `test_single_writer_reasoning_stream_uses_the_writer_session` now fails if single-writer mode ever opens a second session.
- **Not done / not claimed:** CI results on `9799cc7` are pending at the time of writing. No same-host Postgres timing yet.

https://claude.ai/code/session_01DT44mj61Qm3x6KLrsAvEt4


yochze:
`unit-tests` is red here on 7 tests, and none of them come from this PR. I reproduced the same 7 failures on `origin/main` (8683167), which carries none of this PR's changes:

- `tests/unit/test_opus_5_5_support.py::test_opus_5_family_never_gets_budget_tokens` (6 params). a7a8ca7 ("honor reasoning effort…") made `_effort_to_thinking_config` return `{"type": "adaptive", "effort": <effort>}`, and the test still expects `{"type": "adaptive"}`. The anthropic client drops `effort` from `thinking` and sends it as `output_config.effort`, so the test looks stale rather than the code being wrong. Proposed patch: `assert _effort_to_thinking_config(effort, model_id) == {"type": "adaptive", "effort": effort}`.
- `tests/unit/test_sandbox_feedback_loop.py::TestSandboxRulesInPrompts::test_create_data_prompt_lists_forbidden_builtins`: the coder prompt no longer lists the forbidden sandbox builtins. This is also red on main. Which change removed the list still needs to be established.

I found no open fix PR for either. I'm leaving them out of this PR so it doesn't widen.

---

PR #1183, opened 2026-09-23
fix(context): keep revoked columns and unchecked @table mentions out of the agent prompt

yochze:
**What:** the agent's prompt now contains only columns the user can still reach, and an @table mention is rendered only when the run's own schema context would show that table.
**Why:** the per-user column query in the schema builder skipped the `is_accessible` filter, and table mentions rendered whatever table id the client sent, straight from the raw catalog row.

No linked issue. Both problems were found while researching per-column selection in the tables selector.

## Executive summary
The problem affects agents where each user connects with their own database login (`user_required`, including on-behalf-of). A customer's DBA revokes a column, e.g. `customer_ssn`. BOW's own schema view hides it correctly, but the AI is still told the column exists. The AI then writes queries against a column the user can't read, and the column name stays visible in the prompt.

A second path went further. A user could @mention a table in chat (or send a crafted request) and BOW would describe it to the AI in full, even when the admin had not enabled that table on the agent and the user's own login can't access it. In the sandbox, a crafted mention exposed `payroll(employee, salary)` to a user with no grant on it. After this PR the prompt matches what the user is actually allowed to see.

## Product impact
| Scenario | Before | After |
|---|---|---|
| DBA revokes `orders.customer_ssn` from the analyst; analyst asks about orders | Agent is told `orders` has 4 columns, including `customer_ssn` | Agent sees `id, region, amount` |
| Analyst @mentions `orders` from the picker | Mention lists `customer_ssn`, shown as garbled `{'name': ...}:None` strings, with agent name `null` | `id:integer, region:text, amount:numeric` · agent `Shop` |
| Request carries a mention id for `payroll` (inactive on the agent, no grant for the analyst) | Table and columns (`employee`, `salary`) are put in the prompt | Mention is dropped |

## Technical mechanism
```mermaid
flowchart LR
  subgraph Before
    M1[mention object_id from client] --> R1[db.get DataSourceTable] --> P1[raw columns JSON into prompt]:::bad
    O1[UserDataSourceColumn rows] -->|no is_accessible filter| S1[schema section]:::bad
  end
  subgraph After
    M2[mention object_id] --> T2[table on one of the run's agents?] --> B2["SchemaContextBuilder.build(active_only, user)"] --> V2{id in visible tables?}
    V2 -->|yes| P2[columns from resolved table]:::good
    V2 -->|no| D2[dropped]:::good
    O2[UserDataSourceColumn rows] -->|is_accessible = true| S2[schema section]:::good
  end
  classDef bad fill:#fdd,stroke:#c00
  classDef good fill:#dfd,stroke:#080
```

## Reviewer decision box
| | |
|---|---|
| **Risk** | Low–medium. Two filters and one resolution path, all in context building. No schema or API changes. |
| **Blast radius** | Agent prompt only: the schema section on `user_required` overlay connections, and the mentions section. The tables selector and APIs are untouched. |
| **Behavior change (intended)** | Revoked overlay columns leave the prompt. Table mentions for tables the run can't see are dropped instead of rendered. Mention previews are now `name:dtype`. |
| **Verified ✅** | 8 new tests fail on the old code and pass on the new. 153 targeted unit and e2e tests pass. Live sandbox before/after against real Postgres column grants and the real UI mention picker (below). |
| **NOT verified ⚠️** | The full backend unit suite and the `--db=postgres` CI leg (left to CI). A real LLM answer: the sandbox Anthropic key had no credit, so verification reads the saved context snapshot, which is built before the LLM call. |
| **Look first** | `mention_context_builder.py` `_resolve_table_mention` / `_visible_tables`. `schema_context_builder.py`: the overlay column query, `PromptTable(id=...)`, and `split_file_scopes`. |

## Evidence
| Check | Result |
|---|---|
| New regression tests on the unfixed code | 8 failed |
| New regression tests with the fix | 8 passed |
| Targeted context / mention / overlay / schema unit and e2e files (`--db=sqlite`) | **153 passed, 0 failed** |
| Live: rendered schema for `public.orders` | before `cols="4"` incl. `customer_ssn` → after `cols="3"` |
| Live: UI @orders mention | before 4 columns incl. `customer_ssn` → after 3 |
| Live: crafted @payroll mention | before rendered with `salary` → after dropped |
| Failures caused by this PR | none observed |

The live loop, with commands, is in `docs/feedback-loops/revoked-columns-and-table-mentions.md`. Setup: PostgreSQL 16 with `GRANT SELECT (id, region, amount, customer_ssn) ON orders TO analyst`, a `user_required` agent, the admin running as `analyst`, then `REVOKE SELECT (customer_ssn)` and a refresh of the user's schema. The mention was picked in the real UI (`@` → Agents → Shop → `public.orders`), and each result was read from `context_snapshots.context_view_json`.

## Context map
- **Key files**
  - `backend/app/ai/context/builders/schema_context_builder.py`: the overlay column filter; `PromptTable.id` is now set; new `split_file_scopes` flag, default unchanged
  - `backend/app/ai/context/builders/mention_context_builder.py`: table mentions now resolve through the schema builder
  - `backend/app/ai/context/context_hub.py`: passes `data_sources` to the mention builder
  - Tests: `backend/tests/unit/test_mention_table_visibility.py`, plus a parametrized revoked-column test in `backend/tests/unit/test_overlay_inactive_canonical_context.py`
- **Convention:** anything that puts a table or column into the prompt from a client-supplied id should resolve it through `SchemaContextBuilder` for the run's agents and user, never from the raw `DataSourceTable` row. FILE mentions already follow this rule (`MentionService._viewable_mentioned_file_ids`).
- **Follow-ups (not in this PR):**
  - `DATA_SOURCE` mentions still resolve any id without an access check. They only render the agent's name.
  - This groundwork precedes per-column check/uncheck in the tables selector, which will need the same builder-level column filtering.

## Deep detail
**Revoked columns.** `_upsert_user_overlay` keeps revoked column rows with `is_accessible=False`. `read_user_data_source_schema`, which the `/schema` API uses, filters them out, but the builder's overlay query didn't. The fix adds the same predicate.

**Mentions.** The old branch trusted `Mention.object_id`, which `create_completion_mentions` copies from the request's `mentions` payload with no check for tables. The new path:
1. The table must belong to one of the run's agents.
2. `SchemaContextBuilder.build(data_source_ids=[ds], active_only=True, split_file_scopes=False)` runs as the caller.
3. The mention is kept only if its canonical id appears in the result, and its columns come from that result.

This reuses every existing rule rather than re-implementing it: activation, open/overlay/denied connection classification, overlay columns, and unhealthy connections. `split_file_scopes=False` keeps file-connection rows addressable, so file-table mentions still resolve. The build is cached per agent for the run, because `refresh_warm` rebuilds mentions on every agent loop step.

**Road not taken:** matching mentions by table name through `table_names=`. Overlay names can differ from canonical names (Power BI renames), so matching uses the canonical id, which `PromptTable.id` now carries. No other code reads `PromptTable.id` (grep of `app/ai`).

**Self-found during the work:** the first version cleared the cache on every `build()`. I switched it to per-run after seeing that `refresh_warm` runs inside the agent loop.

https://claude.ai/code/session_01JkLvF8xsbDibmrdTYB7YJi

---
_Generated by [Claude Code](https://claude.ai/code/session_01JkLvF8xsbDibmrdTYB7YJi)_


yochze:
**`unit-tests` is red for reasons unrelated to this PR.** It fails with 7 failed and 5546 passed. The same 7 tests fail identically on the base commit `8683167` (current `main`). I reproduced that locally on a clean checkout of `8683167`: `7 failed, 12 passed` across the two files below. This PR doesn't touch either area, and I found no open PR that fixes them. #1181's `unit-tests` job is also red.

1. `tests/unit/test_opus_5_5_support.py::test_opus_5_family_never_gets_budget_tokens[*]` (6 cases)
   - Since `a7a8ca7` ("honor reasoning effort"), `_effort_to_thinking_config` returns `{"type": "adaptive", "effort": e}`. `effort` is metadata that adapters drop before calling the API (see the comment in `app/ai/llm/reasoning.py`).
   - The test still asserts exact equality with `{"type": "adaptive"}`, so the test is stale.
   - Proposed patch:
     ```python
     cfg = _effort_to_thinking_config(effort, model_id)
     assert cfg["type"] == "adaptive" and "budget_tokens" not in cfg
     ```
2. `tests/unit/test_sandbox_feedback_loop.py::TestSandboxRulesInPrompts::test_create_data_prompt_lists_forbidden_builtins`
   - `c727020` ("cut agent-loop token cost via prompt-cache fixes") moved the sandbox rules into the codegen **system** prompt.
   - The test's fake LLM records only `messages[0].content`, the user half, so it no longer sees the rules. This is also a stale test.
   - Proposed patch: have the fake's `inference_stream_v2` record `kwargs.get("system")` together with the message content, and assert on both.

I'm not widening this PR with these. I tried re-running the failed job once, but the integration isn't allowed to (`403 Resource not accessible by integration`). Since the failure reproduces deterministically on `main`, a re-run would fail the same way anyway. The PR stays watched.

---

